PR #13, opened 2019-11-04
Add for.jl

cscherrer:
This PR adds the `For` combinator from Soss.jl. I usually use it like this:

```julia
x ~ For(indices) do j
    someDistribution(j)
end
```

`indices::T` can have...

- `T <: NTuple{N,J} where {J <: Integer}` 
- `T <: NTuple{N,J} where {J <: AbstractUnitRange}`
- `T <: Base.Generator` (for example to use with `eachrow`)

For a single index, there's no need to write this an a 1-tuple, that conversion is done automatically

### DIFF
--- a/src/for.jl
+++ b/src/for.jl
@@ -1,0 +1,97 @@
+using Distributions
+import Distributions.logpdf
+using Base.Cartesian
+
+
+export logpdf
+export rand
+
+export For
+struct For{F,T,D,X} 
+    f :: F  
+    θ :: T
+end
+
+#########################################################
+# T <: NTuple{N,J} where {J <: Integer}
+#########################################################
+
+For(f, θ::J...) where {J <: Integer} = For(f,θ)
+
+function For(f::F, θ::T) where {F, N, J <: Integer, T <: NTuple{N,J}}
+    d = f.(ones(Int, N)...)
+    D = typeof(d)
+    X = eltype(d)
+    For{F, NTuple{N,J}, D, X}(f,θ)
+end
+
+@inline function logpdf(d::For{F,T,D,X1},xs::AbstractArray{X2,N}) where {F, N, J <: Integer, T <: NTuple{N,J}, D,  X1,  X2 <: X1}
+    s = 0.0
+    @inbounds @simd for θ in CartesianIndices(d.θ)
+        s += logpdf(d.f(Tuple(θ)...), xs[θ])
+    end
+    s
+end
+
+function Base.rand(dist::For) 
+    map(CartesianIndices(dist.θ)) do I
+        (rand ∘ dist.f)(Tuple(I)...)
+    end
+end
+
+#########################################################
+# T <: NTuple{N,J} where {J <: AbstractUnitRange}
+#########################################################
+
+For(f, θ::J...) where {J <: AbstractUnitRange} = For(f,θ)
+
+function For(f::F, θ::T) where {F, N, J <: AbstractRange, T <: NTuple{N,J}}
+    d = f.(ones(Int, N)...)
+    D = typeof(d)
+    X = eltype(d)
+    For{F, NTuple{N,J}, D, X}(f,θ)
+end
+
+
+@inline function logpdf(d::For{F,T,D,X1},xs::AbstractArray{X2,N}) where {F, N, J <: AbstractRange,  T <: NTuple{N,J}, D, X1, X2 <: X1}
+    s = 0.0
+    @inbounds @simd for θ in CartesianIndices(d.θ)
+        s += logpdf(d.f(Tuple(θ)...), xs[θ])
+    end
+    s
+end
+
+
+function Base.rand(dist::For{F,T}) where {F,  N, J <: AbstractRange, T <: NTuple{N,J}}
+    map(CartesianIndices(dist.θ)) do I
+        (rand ∘ dist.f)(Tuple(I)...)
+    end
+end
+
+#########################################################
+# T <: Base.Generator
+#########################################################
+
+function For(f::F, θ::T) where {F, T <: Base.Generator}
+    d = f(θ.f(θ.iter[1]))
+    D = typeof(d)
+    X = eltype(d)
+    For{F, T, D, X}(f,θ)
+end
+
+
+@inline function logpdf(d :: For{F,T}, x) where {F,T <: Base.Generator}
+    s = 0.0
+    for (θj, xj) in zip(d.θ, x)
+        s += logpdf(d.f(θj), xj)
+    end
+    s
+end
+
+@inline function rand(d :: For{F,T,D,X}) where {F,T <: Base.Generator, D, X}
+    rand.(Base.Generator(d.θ.f, d.θ.iter))
+end
+
+#########################################################
+
+


yebai:
Many thanks, @cscherrer - could you add a simple test for the new feature?

mohamed82008:
Thanks @cscherrer, I will take a closer look at this when I get a chance, having a hectic week.

src/for.jl
```diff
@@ -0,0 +1,97 @@
...
+    d = f.(ones(Int, N)...)
+    D = typeof(d)
+    X = eltype(d)
+    For{F, NTuple{N,J}, D, X}(f,θ)
+end
+
+@inline function logpdf(d::For{F,T,D,X1},xs::AbstractArray{X2,N}) where {F, N, J <: Integer, T <: NTuple{N,J}, D,  X1,  X2 <: X1}
+    s = 0.0
+    @inbounds @simd for θ in CartesianIndices(d.θ)
+        s += logpdf(d.f(Tuple(θ)...), xs[θ])
```
mohamed82008: This is not ideal for Tracker and GPU support. The better version here is:
```julia
g(cind) = f(Tuple(cind)...)
inds = CartesianIndices(d.θ)
s = sum(logpdf.(g.(inds), xs))
```
This will be slower on the CPU and when using `ForwardDiff` but for `Tracker` it will be more efficient.

```julia
function testf1(x)
    f(i, j) = Normal(i + j)
    θ = (10, 10)
    g(cind) = f(Tuple(cind)...)
    inds = CartesianIndices(θ)
    s = sum(logpdf.(g.(inds), x))
    return s
end

function testf2(x)
    f(i, j) = Normal(i + j)
    _θ = (10, 10)
    s = 0.0
    @inbounds @simd for θ in CartesianIndices(_θ)
        s += logpdf(f(Tuple(θ)...), x[θ])
    end
    return s
end
x = rand(10, 10)
xparam = param(x)

using BenchmarkTools

@btime testf1($x); # 714 ns
@btime testf2($x); # 634 ns

@btime testf1($xparam); # 1.25 μs
@btime testf2($xparam); #14.6 μs
```

src/for.jl
```diff
@@ -0,0 +1,97 @@
...
+end
+
+#########################################################
+# T <: NTuple{N,J} where {J <: Integer}
+#########################################################
+
+For(f, θ::J...) where {J <: Integer} = For(f,θ)
+
+function For(f::F, θ::T) where {F, N, J <: Integer, T <: NTuple{N,J}}
+    d = f.(ones(Int, N)...)
```
mohamed82008: What if different indices return different distributions?

mohamed82008: We can check this without running the code using `Core.Compiler.return_type`, e.g. `Core.Compiler.return_type(+, Tuple{Int, Int}) === Int`. If inference fails then we can assume the output is anything.

src/for.jl
```diff
@@ -0,0 +1,97 @@
...
+
+function For(f::F, θ::T) where {F, N, J <: AbstractRange, T <: NTuple{N,J}}
+    d = f.(ones(Int, N)...)
+    D = typeof(d)
+    X = eltype(d)
+    For{F, NTuple{N,J}, D, X}(f,θ)
+end
+
+
+@inline function logpdf(d::For{F,T,D,X1},xs::AbstractArray{X2,N}) where {F, N, J <: AbstractRange,  T <: NTuple{N,J}, D, X1, X2 <: X1}
```
mohamed82008: This is identical to the previous one? We can do `J <: Union{Integer, AbstractRange}` and combine them.

src/for.jl
```diff
@@ -0,0 +1,97 @@
...
+    s = 0.0
+    @inbounds @simd for θ in CartesianIndices(d.θ)
+        s += logpdf(d.f(Tuple(θ)...), xs[θ])
+    end
+    s
+end
+
+
+function Base.rand(dist::For{F,T}) where {F,  N, J <: AbstractRange, T <: NTuple{N,J}}
```
mohamed82008: Same as previous method?

src/for.jl
```diff
@@ -0,0 +1,97 @@
...
+    end
+end
+
+#########################################################
+# T <: NTuple{N,J} where {J <: AbstractUnitRange}
+#########################################################
+
+For(f, θ::J...) where {J <: AbstractUnitRange} = For(f,θ)
+
+function For(f::F, θ::T) where {F, N, J <: AbstractRange, T <: NTuple{N,J}}
```
mohamed82008: Same constructor as previous one for `J <: Integer`? What's the difference between these 2?

src/for.jl
```diff
@@ -0,0 +1,97 @@
...
+
+function For(f::F, θ::T) where {F, T <: Base.Generator}
+    d = f(θ.f(θ.iter[1]))
+    D = typeof(d)
+    X = eltype(d)
+    For{F, T, D, X}(f,θ)
+end
+
+
+@inline function logpdf(d :: For{F,T}, x) where {F,T <: Base.Generator}
```
mohamed82008: Please use broadcasting here as well for Tracker and GPU support.

src/for.jl
cscherrer: The assumption here is that this isn't the case, in order to avoid dynamic dispatch. My suggestion would be for the user to handle multiple return distributions by instead always returning a mixture.

But I'd never seen `Core.Compiler.return_type`, that looks really useful!

src/for.jl
cscherrer: Moderate hit on floats in exchange for great AD benefits, works for me. IIUC `Tracker` is here for now but will go away as Zygote matures, any thoughts how this will effect the comparison?

Also, I've played with some options for doing this with a parallel reduction, i.e. thread-local on a CPU. Maybe this should be a different combinator, maybe `ParFor`? Or as a new argument could be ok too, whatever is best for users without sacrificing performance

mohamed82008: Zygote promises to support loops like this efficiently, but in my experience Zygote is still not stable enough to adopt in a relatively mature project like Turing. The parallel stuff in Julia 1.3 also looks very interesting. I still haven't experimented with it though, so I will have to do that before commenting any further.

src/for.jl
mohamed82008: I think with `Core.Compiler.return_type`, one can support the single and multiple return distribution types simultaneously. The former will be inferred and use static dispatch while the latter provides an additional feature for a performance hit which some users might still want in the experimental phase of their project before optimizing the code's performance.

src/for.jl
cscherrer: For `Tracker` support, could it make more sense to just specify the adjoint separately?

There's some discussion of parallel reduction in this context here:
https://discourse.julialang.org/t/parallel-reductions/30180/8

src/for.jl
cscherrer: This is great in principle, but it's pretty strange that there's no multi-argument method:

```julia
julia> methods(Core.Compiler.return_type)
# 2 methods for generic function "return_type":
[1] return_type(f, t) in Core.Compiler at compiler/typeinfer.jl:626
[2] return_type(f, t, world) in Core.Compiler at compiler/typeinfer.jl:631
```

Maybe I'm missing something?

cscherrer:
Thanks for your review @mohamed82008. For your question about the methods, my implementation may need adjustments, but I think I should clarify the intent.

The `For` constructor is defined as
```julia
struct For{F,T,D,X}
    f :: F
    θ :: T
end
```

The important cases I see to support are

```julia
# T <: Tuple{Int}
For(5,2) do i,j ... end
```

```julia
# T <: Base.OneTo
For(eachindex(x)) do xj ... end
```

```julia
# T <: Base.Generator
For(eachrow(X) do xrow ... end
```

Supertypes of the second and third cases are
```julia
# `tower` is my little helper function

julia> tower(eachrow(rand(100,3)))
2-element Array{DataType,1}:
 Base.Generator{Base.OneTo{Int64},Base.var"#174#175"{Array{Float64,2}}}
 Any                                                                   

julia> tower(eachindex(rand(100)))
6-element Array{DataType,1}:
 Base.OneTo{Int64}        
 AbstractUnitRange{Int64} 
 OrdinalRange{Int64,Int64}
 AbstractRange{Int64}     
 AbstractArray{Int64,1}   
 Any  
```

I think the big question is, what's the most general we can make these without sacrificing performance and predictable behavior?

src/for.jl
mohamed82008: I mean it's possible but getting it to perform better than the baseline implementation with broadcasting might not be easy.

src/for.jl
mohamed82008: The second argument to `return_type` goes `Tuple{arg1T, arg2T, ...}`, see my `+` example above.

mohamed82008:
I will experiment with a slightly different implementation.

yebai:
Since this PR does not interact with other parts of DistributionsAD, I'll merge this PR for now; further improvements can be implemented in separate PRs.

yebai:
Many thanks to @cscherrer!

mohamed82008:
Cool! Sorry @cscherrer for not coming back to this earlier, got busy with other stuff.

PR #15, opened 2019-11-09
Add compatibilities and bump patch version

torfjelde:
Depends on the following PRs:
- [x] https://github.com/FluxML/Tracker.jl/pull/55
- [x] https://github.com/FluxML/Zygote.jl/pull/398

`DistributionsAD` currently breaks for newer versions of `Zygote` and other related/dependent packages, including `Turing`. The lower-bounds in this PR provide a fix.

UPDATE: both PRs have been merged, now we're just waiting for relases to happen.

### DIFF
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -2,9 +2,9 @@
 
 [[AbstractFFTs]]
 deps = ["LinearAlgebra"]
-git-tree-sha1 = "380e36c66edfa099cd90116b24c1ce8cafccac40"
+git-tree-sha1 = "051c95d6836228d120f5f4b984dd5aba1624f716"
 uuid = "621f4979-c628-5d54-868e-fcf4e3e8185c"
-version = "0.4.1"
+version = "0.5.0"
 
 [[Adapt]]
 deps = ["LinearAlgebra"]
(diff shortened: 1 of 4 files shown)


torfjelde:
It seems lie builds are failing because the Travis build isn't using the most up-to-date package registry. Locally, doing `] up` results in `Zygote@0.4.1` while in the Travis log you can see `Zygote@0.3.4` is being used.

mohamed82008:
Please update the Manifest file.

torfjelde:
> Please update the Manifest file.

Hah, just edited my comment saying the same thing. You want me to update or remove the Manifest.toml?

mohamed82008:
Hmm, I would rather keep it to be able to easily reproduce Travis failures when they happen, but it's a matter of taste.

torfjelde:
Seems like good reasoning to me :+1: Updated and tests are running now. Should pass.

Worth noting that the version-bounds I've added, especially for Zygote. The reason for this is the breaking changes introduced in `Zygote@0.4`.

mohamed82008:
Cool thanks for this PR!

torfjelde:
This should be ready now:)

mohamed82008:
Thanks @torfjelde !

PR #20, opened 2020-01-10
Fix method ambiguity

devmotion:
Fixes the method ambiguity errors seen in Turing's CI tests (see, e.g., https://github.com/TuringLang/Turing.jl/pull/1063).

BTW I had some issues reproducing the error locally after skipping some other tests - `import/using`s and definitions/settings of tests that are run prior to the distribution tests affect them. Maybe one could think about using [SafeTestsets](https://github.com/YingboMa/SafeTestsets.jl) to avoid such issues.

### DIFF
--- a/src/common.jl
+++ b/src/common.jl
@@ -115,6 +115,4 @@ end
 ## PDMats ##
 
 PDMats.invquad(Σ::PDiagMat, x::Tracker.TrackedVector) = sum(abs2.(x) ./ Σ.diag)
-getchol(m::PDMats.AbstractPDMat) = m.chol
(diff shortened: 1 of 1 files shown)


devmotion:
Maybe it would be good if someone else could check locally that this PR actually fixes the test errors in Turing (such as https://github.com/TuringLang/Turing.jl/pull/1063, https://github.com/TuringLang/Turing.jl/pull/1064, https://github.com/TuringLang/Turing.jl/pull/1067, and https://github.com/TuringLang/Turing.jl/pull/1069).

mohamed82008:
I will check this tomorrow.

mohamed82008:
Thanks @devmotion !

PR #23, opened 2020-01-15
Support Distributions 0.22

mohamed82008:


devmotion:
Unfortunately, due to lines such as https://github.com/TuringLang/DistributionsAD.jl/blob/ab14b91e723340091359cce535c223ad884a540c/src/multivariate.jl#L89-L91 the breaking change introduced in Distributions 0.22 will lead to different behaviour of MvNormal constructors in Distributions and DistributionsAD. I don't really see how one could support both Distributions 0.21 and 0.22, so I guess the way forward would be to fix the implementation of MvNormal in DistributionsAD and drop support for Distributions 0.21.

mohamed82008:
Sounds good.

mohamed82008:
Btw IIUC, the different behavior is not really an issue in this case. For one, we are not committing any additional type piracy now than we were earlier. It's just going to use a diagonal covariance for `UniformScaling` here when `Distributions` is using a `ScalMat`, is this about right? I can fix it to use a `ScalMat`-like implementation here instead but just checking that this is the only "problem".

mohamed82008:
Oh I see the problem, it's the squaring, sorry.

src/multivariate.jl
```diff
@@ -1,52 +1,82 @@
...
 
-function _logpdf(d::TuringDiagNormal, x::AbstractVector)
-    return -(dim(d) * log(2π) + 2 * sum(log.(d.σ)) + sum(abs2, (x .- d.m) ./ d.σ)) / 2
+function _logpdf(d::TuringScalMvNormal, x::AbstractVector)
+    return -(length(x) * log(2π * abs2(d.σ)) + sum(abs2, (x .- d.m) ./ d.σ)) / 2
 end
-function _logpdf(d::TuringDiagNormal, x::AbstractMatrix)
-    return -(dim(d) * log(2π) .+ 2 * sum(log.(d.σ)) .+ sum(abs2, (x .- d.m) ./ d.σ, dims=1)') ./ 2
+function _logpdf(d::TuringScalMvNormal, x::AbstractMatrix)
+    return -(size(x, 2) * log(2π) .+ 2 * sum(log(d.σ)) .+ sum(abs2, (x .- d.m) ./ d.σ, dims=1)') ./ 2
```
devmotion: This seems wrong, I think it should be
```suggestion
    return -(size(x, 2) * (log(2π) + 2 * log(d.σ)) .+ sum(abs2, (x .- d.m) ./ d.σ, dims=1)') ./ 2
```
or
```suggestion
    return -(size(x, 2) * log(2π * abs2(d.σ)) .+ sum(abs2, (x .- d.m) ./ d.σ, dims=1)') ./ 2
```
as in the vector-valued case above. IMO it's a bit surprising to use `size(x, 2)` and `length(x)` instead of `length(d)`, but it doesn't matter I guess.

I'm also wondering if you actually want to use `sum(abs2, ....)'` since that will just transpose the vector but keep both dimensions
```julia
julia> sum(abs2, rand(10, 20); dims = 1)'
20×1 LinearAlgebra.Adjoint{Float64,Array{Float64,2}}:
```
Hence the result is not a `Vector{<:Real}` even though the default implementation in Distributions returns a vector:

https://github.com/JuliaStats/Distributions.jl/blob/70f3d97fca49b0abfa154f2f29c7b4f4a90b2ff8/src/multivariates.jl#L230-L235

devmotion: I guess you want to use
```suggestion
    return -(size(x, 2) * (log(2π) + 2 * log(d.σ)) .+ vec(sum(abs2, (x .- d.m) ./ d.σ, dims=1))) ./ 2
```

devmotion: Or maybe even
```suggestion
    return -(size(x, 2) * (log(2π) + 2 * log(d.σ)) .+ vec(sum(abs2, x .- d.m, dims=1)) ./ abs2(d.σ)) ./ 2
```

src/multivariate.jl
```diff
@@ -1,52 +1,82 @@
...
 end
-function _logpdf(d::TuringMvNormal, x::AbstractVector)
-    return -(dim(d) * log(2π) + logdet(d.C) + sum(abs2, zygote_ldiv(d.C.U', x .- d.m))) / 2
+
+function _logpdf(d::TuringDiagMvNormal, x::AbstractVector)
+    return -(length(x) * log(2π) + 2 * sum(log.(d.σ)) + sum(abs2, (x .- d.m) ./ d.σ)) / 2
```
devmotion:
```suggestion
    return -(length(x) * log(2π) + 2 * sum(log, d.σ) + sum(abs2, (x .- d.m) ./ d.σ)) / 2
```

src/multivariate.jl
```diff
@@ -1,52 +1,82 @@
...
+end
+function _logpdf(d::TuringDiagMvNormal, x::AbstractMatrix)
+    return -(size(x, 2) * log(2π) .+ 2 * sum(log.(d.σ)) .+ sum(abs2, (x .- d.m) ./ d.σ, dims=1)') ./ 2
```
devmotion:
```suggestion
    return -(size(x, 2) * log(2π) .+ 2 * sum(log, d.σ) .+ vec(sum(abs2, (x .- d.m) ./ d.σ, dims=1))) ./ 2
```

src/multivariate.jl
mohamed82008: Good catch. Thanks. I will fix and test the matrix case.

src/multivariate.jl
mohamed82008: The broadcasting versions is actually a little faster with `TrackedVector` but the opposite is true for normal vectors.

src/multivariate.jl
```diff
@@ -75,20 +105,26 @@ function MvNormal(
...
-MvNormal(m::TrackedVector{<:Real}, A::UniformScaling{<:Real}) = MvNormal(m, A.λ)
+function MvNormal(m::TrackedVector{<:Real}, A::UniformScaling{<:TrackedReal})
+    return MvNormal(m, sqrt(A.λ))
+end
+function MvNormal(m::AbstractVector{<:Real}, A::UniformScaling{<:TrackedReal})
+    return MvNormal(m, sqrt(A.λ))
+end
+function MvNormal(m::TrackedVector{<:Real}, A::UniformScaling{<:Real})
+    return MvNormal(m, sqrt(A.λ))
+end
```
devmotion: I think the three constructors above that create an `MvNormal` instance might not be needed anymore. Hopefully that just works with the updated constructors in Distributions 0.22.

src/multivariate.jl
```diff
@@ -75,20 +105,26 @@ function MvNormal(
...
 
 # zero mean,, constant variance
 MvNormal(d::Int, σ::TrackedReal{<:Real}) = MvNormal(zeros(d), σ)
```
devmotion: This line is still needed if one does not want to track the mean as well (the default behaviour in Distributions is to create a mean vector `Zeros{typeof(σ)}(d)`
https://github.com/JuliaStats/Distributions.jl/blob/70f3d97fca49b0abfa154f2f29c7b4f4a90b2ff8/src/multivariate/mvnormal.jl#L221

src/multivariate.jl
mohamed82008: I think it's needed to avoid calling `ScalMat` in https://github.com/JuliaStats/Distributions.jl/blob/70f3d97fca49b0abfa154f2f29c7b4f4a90b2ff8/src/multivariate/mvnormal.jl#L208. We want `Tracker`-`MvNormal` calls to have nothing to do with `PDMats` to simplify things.

devmotion: Ah sorry, I missed that of course you're just forwarding to the specialized implementation for TrackedReal but never call the constructors in Distributions.

mohamed82008:
So while testing the matrix case, I found some other bugs and fixed them including a performance bug that was probably the reason why `Multi` of `MvNormal` was slow, (CC: @xukai92). I also found 2 failed tests cases for `MvNormal` which I don't know why yet. You can find them in `broken_mult_cont_dists` in runtests.jl.

mohamed82008:
Here is a MWE of a failed test case if you want to investigate:
```julia
using Distributions, DistributionsAD, ForwardDiff, Tracker
function f(x)
	@assert length(x) == 8
	C = reshape(x[1:4], 2, 2)
	val = reshape(x[5:8], 2, 2)
	return sum(logpdf(MvNormal(C), val))
end
x = [1.0, 0.0, 0.0, 1.0]; x = [x; ones(4)];
Tracker.data(Tracker.gradient(f, x)[1])
ForwardDiff.gradient(f, x)
```

mohamed82008:
I will add more tests to raise the coverage, then if there are no objections after that, I will merge. We can look into the test failures later if we can't figure it out soon.

devmotion:
I don't have time right now to look into these test failures, but since the main motivation for this PR initially was to support Distributions 0.22, I think it can be merged.

As a general comment, I'm not sure if one should actually define three different Tracker-compatible Gaussian distributions, it somehow doesn't feel "right". It feels like they differ just in the type/structure of the covariance matrix, and hence IMO the implementation should also handle this on the level of the covariance matrices, similar to how `MvNormal` supports `ScalMat`, `PDiagMat`, and `PDMat` as covariance matrices and `FillArray` as mean vector. But also that could/should be part of a separate PR IMO.

mohamed82008:
> As a general comment, I'm not sure if one should actually define three different Tracker-compatible Gaussian distributions, it somehow doesn't feel "right".

I agree, let's do that in another PR.

devmotion:
Regarding the incorrect gradients @mohamed82008 mentioned in his comment above, the problem with ForwardDiff seems to be that basically all linear algebra implementations that exploit the sparsity of matrices (by, e.g., calling `convert(UpperTriangular, ...)`) are AD unsafe (see, e.g., https://github.com/JuliaDiff/ForwardDiff.jl/issues/197#issuecomment-516003182 and the other comments in this issue). For instance you can observe this in the following small example, which could be related to the problem mentioned above:
```julia
using ForwardDiff
using LinearAlgebra

@show ForwardDiff.gradient(x -> sum(x \ ones(2)), Matrix(I, 2, 2))
@show ForwardDiff.gradient(x -> sum(cholesky(x) \ ones(2)), Matrix(I, 2, 2))
```

PR #25, opened 2020-01-18
Zygote support

mohamed82008:
This PR makes most currently working distributions `Zygote`-compatible.


mohamed82008:
This PR requires https://github.com/FluxML/Zygote.jl/pull/469 to be merged first.

src/matrixvariate.jl
```diff
@@ -0,0 +1,218 @@
...
+    p = size(C, 1)
+    h_df * (logdet(C) + p * typeof(df)(logtwo)) + logmvgamma(p, h_df)
+end
+
+#### Properties
+
+Distributions.insupport(::Type{TuringWishart}, X::Matrix) = isposdef(X)
+Distributions.insupport(d::TuringWishart, X::Matrix) = size(X) == size(d) && isposdef(X)
+
+PDMats.dim(d::TuringWishart) = size(d.chol, 1)
```
devmotion: That should not be needed? The interface in Distributions only uses `size` and `length` (and with https://github.com/TuringLang/Turing.jl/pull/1076 it's not used in Turing anymore).
```suggestion
```

src/matrixvariate.jl
```diff
@@ -0,0 +1,218 @@
...
+    p = size(C, 1)
+    h_df * (p * typeof(df)(logtwo) - logdet(C)) + logmvgamma(p, h_df)
+end
+
+#### Properties
+
+Distributions.insupport(::Type{TuringInverseWishart}, X::Matrix) = isposdef(X)
+Distributions.insupport(d::TuringInverseWishart, X::Matrix) = size(X) == size(d) && isposdef(X)
+
+PDMats.dim(d::TuringInverseWishart) = size(d.S, 1)
```
devmotion: Same here.
```suggestion
```

src/matrixvariate.jl
mohamed82008: It's not needed outside but it's used in the same file in many places.

src/multivariate.jl
```diff
@@ -159,57 +178,80 @@ function _logpdf(d::TuringMvLogNormal, x::AbstractMatrix{T}) where {T<:Real}
...
+    return TuringMvLogNormal(TuringMvNormal(m, A))
 end
 
 # zero mean,, constant variance
-MvLogNormal(d::Int, σ::TrackedReal{<:Real}) = MvLogNormal(zeros(d), σ)
+MvLogNormal(d::Int, σ::TrackedReal{<:Real}) = TuringMvLogNormal(TuringMvNormal(d, σ))
+
+## Zygote adjoint
+
+Zygote.@adjoint function Distributions.MvNormal(
```
devmotion: Using `@adjoint` does not require Zygote but only ZygoteRules in principle, but I'm not sure how much we rely on Zygote-only functions in other places.

src/matrixvariate.jl
devmotion: Yes but one should really just use `length` there.

src/multivariate.jl
mohamed82008: I think we can depend on ZygoteRules only in the src, but still load Zygote anyways to support Tracker + MvNormal automatically without the users having to load Zygote themselves which would be weird.

src/matrixvariate.jl
mohamed82008: Doesn't have to be `PDMats.dim` though, can be just a local `dim` function. `length` can also be defined.

devmotion: Sure, that would be fine. It just feels weird to define a function that does not seem to be part of the official Distributions interface and is taken from PDMats, even though we do not work with `AbstractPDMat`s.

mohamed82008:
Distributions to fix/test:
1. Beta
2. Chernoff
3. Chisq
4. Erlang
5. Gamma
6. FDist
7. TDist
8. Wishart
9. InverseWishart
10. MatrixBeta

src/univariate.jl
```diff
@@ -1,16 +1,50 @@
...
+    l = -log(diff)
+    f = isfinite(l)
+    da = 1/diff
+    n = T(NaN)
+    return l, Δ->(f ? da : n, f ? -da : n, f ? zero(T) : n)
+end
+Zygote.@adjoint function Distributions.Uniform(args...)
+    value, back = Zygote.pullback(TuringUniform, args...)
+    return value, x -> back(x)
+end
```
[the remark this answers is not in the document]
mohamed82008: Neat!

src/univariate.jl
```diff
@@ -1,16 +1,50 @@
...
+    diff = Tracker.data(b) - Tracker.data(a)
+    T = typeof(diff)
+    l = -log(diff)
     f = isfinite(l)
-    da = 1/(b - a)
+    da = 1/diff
     n = T(NaN)
     return l, Δ->(f ? da : n, f ? -da : n, f ? zero(T) : n)
 end
+Zygote.@adjoint function uniformlogpdf(a, b, x)
```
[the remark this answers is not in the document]
mohamed82008: Zygote returns `nothing` for `x`'s pullback. When Tracker returned `nothing`, it was causing errors in Turing. So I fixed it then, and I applied the same fix for Zygote to make them consistent. There might be a better solution here. I can come back to it later.

src/multivariate.jl
```diff
@@ -151,65 +171,88 @@ function _logpdf(d::TuringMvLogNormal, x::AbstractVector{T}) where {T<:Real}
     return insupport(d, x) ? (_logpdf(d.normal, log.(x)) - sum(log.(x))) : -T(Inf)
 end
 function _logpdf(d::TuringMvLogNormal, x::AbstractMatrix{T}) where {T<:Real}
-    if all(insupport(d, x))
+    if all(i -> insupport(d, x[:, i]), 1:size(x, 2))
```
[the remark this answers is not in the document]
mohamed82008: Fair point, I will fix it.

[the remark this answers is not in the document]
mohamed82008: The type stability issue is more involved. I will come back to it later. We probably need to use `Core.Compiler.return_type`.

src/matrixvariate.jl
```diff
@@ -0,0 +1,216 @@
...
+
+#### Sampling
+
+Distributions._rand!(rng::AbstractRNG, d::TuringInverseWishart, A::AbstractMatrix) =
+    (A .= inv(cholesky!(_rand!(rng, TuringWishart(d.df, inv(cholesky(d.S))), A))))
+
+## Adjoints
+
+ZygoteRules.@adjoint function Distributions.Wishart(df::Real, S::AbstractMatrix{<:Real})
+    return ZygoteRules.pullback((df, S) -> TuringWishart(df, S), df, S)
```
devmotion:
```suggestion
    return ZygoteRules.pullback(TuringWishart, df, S)
```

src/matrixvariate.jl
```diff
@@ -0,0 +1,216 @@
...
+end
+ZygoteRules.@adjoint function Distributions.InverseWishart(df::Real, S::AbstractMatrix{<:Real})
+    return ZygoteRules.pullback((df, S) -> TuringInverseWishart(df, S), df, S)
```
devmotion:
```suggestion
    return ZygoteRules.pullback(TuringInverseWishart, df, S)
```

src/multivariate.jl
```diff
@@ -151,65 +171,92 @@ function _logpdf(d::TuringMvLogNormal, x::AbstractVector{T}) where {T<:Real}
...
+    insupp = map(1:size(x, 2)) do i 
+        insupport(d, x[:, i])
+    end
+    if all(insupp)
         return _logpdf(d.normal, log.(x)) - vec(sum(log.(x), dims=1))
     else
-        return fill(-T(Inf), size(x, 2))
+        return map(1:size(x, 2)) do i
+            insupp[i] && return -T(Inf)
+            return _logpdf(d.normal, log.(x[:,i])) - sum(log.(x[:,i]))
```
devmotion: I guess that could be simplified to just `map(_logpdf(d, view(x, :, i)), 1:size(x, 2))` or something similar.

src/multivariate.jl
```diff
@@ -151,65 +171,92 @@ function _logpdf(d::TuringMvLogNormal, x::AbstractVector{T}) where {T<:Real}
     return insupport(d, x) ? (_logpdf(d.normal, log.(x)) - sum(log.(x))) : -T(Inf)
 end
 function _logpdf(d::TuringMvLogNormal, x::AbstractMatrix{T}) where {T<:Real}
-    if all(insupport(d, x))
+    insupp = map(1:size(x, 2)) do i 
+        insupport(d, x[:, i])
+    end
```
devmotion: Is it actually necessary to allocate an array? I guess one could just check `if all(i -> insupport(d, view(x, :, i)), 1:size(x, 2))` or something similar.

src/multivariate.jl
mohamed82008: Hmm this will be computationally more expensive, but it solves the type stability.

mohamed82008: Actually it doesn't work because we need to take `log` inside which fails when `x` is not in the support. So we have to return `-Inf` separately.

src/multivariate.jl
mohamed82008: We need `insupp` later in case not all are in the support.

src/multivariate.jl
```diff
@@ -177,6 +177,7 @@ function _logpdf(d::TuringMvLogNormal, x::AbstractMatrix{T}) where {T<:Real}
     if all(insupp)
         return _logpdf(d.normal, log.(x)) - vec(sum(log.(x), dims=1))
     else
+        T = typeof(log(abs(x[1])))
```
mohamed82008: @devmotion this should solve most type stability problems here, right?

src/multivariate.jl
devmotion: I don't see why that would not work if a column of `x` is not in the support - we would just call
https://github.com/TuringLang/DistributionsAD.jl/blob/2ac40ba38f78662e57a91399bac2b7f711cac4d3/src/multivariate.jl#L170-L172
for every column. So as long as the logpdf definition for vectors is type-stable, the matrix-valued version should be trivial? Or am I missing something here?

So IMO the type-stability issue should be handled in the implementation for vectors, and then the definition for matrices could be a simple map.

mohamed82008: Oh I see, ok. Yes that makes sense. There are still 2 issues though:
1. `insupport` would be called twice, once in the matrix version to find `insupp` and another in the vector version.
2. The vector version has the same `T(Inf)` that may not be type stable or even worse may error if `T` is `Int`.

I think the current implementation avoids the double calling of `insupport` and handles the `Int` case and type stability nicely.

devmotion: That's not exactly what I had in mind. I thought you could just define
```julia
function _logpdf(d::TuringMvLogNormal, x::AbstractMatrix{<:Real}) 
    return map(i -> _logpdf(d, view(x, :, i)), axes(x, 2))
end
```
and move any potential fix for the type instability to the definition of `_logpdf(::TuringMvLogNormal, x::AbstractVector{<:Real})`. As long as the implementation for vectors is type stable, there is no type instability, and additionally there is not double calling of `insupport`.

mohamed82008: This will be slower than the matrix version when all columns are in the support and it allocates for the views.

mohamed82008: Well view allocation is still better than `x[:,i]`.

mohamed82008:
```julia
julia> dist = TuringMvLogNormal(MvNormal(rand(10), rand(10)));

julia> x = rand(dist, 100);

julia> f(dist, x) = logpdf(dist, x);

julia> g(dist, x) = [logpdf(dist, view(x, :, i)) for i in 1:size(x,2)];

julia> @btime f($dist, $x);
  18.800 μs (17 allocations: 29.63 KiB)

julia> @btime g($dist, $x);
  35.700 μs (503 allocations: 68.13 KiB)
```

mohamed82008: I will mark this as resolved. Feel free to unmark it if you think there is a better way to do this.

yebai:
@mohamed82008 I reordered the todo list, according to how often each distribution may be used. I think we can focus on the top list, and come back to the remaining in a separate PR.

mohamed82008:
Sounds good. I can probably wrap this PR up tomorrow.

src/multivariate.jl
devmotion: While performing some benchmarks I noticed that the current implementation is incorrect - it throws an error if samples are not in the support due to the line `insupp[i] && return -T(Inf)` which should be `insupp[i] || return -T(Inf)`.

I performed the following benchmarks (after fixing this bug but without fixing any type instabilities in `_logpdf(d::TuringMvLogNormal, x::AbstractVector)` since for the benchmarked types there should not be any issues):
```julia
using DistributionsAD
using BenchmarkTools

dist = TuringMvLogNormal(TuringDiagMvNormal(rand(10), rand(10)))

x = rand(dist, 100)
y = copy(x)
y[3, 20] = -1

f(dist, x) = DistributionsAD._logpdf(dist, x)
g(dist, x) = [DistributionsAD._logpdf(dist, view(x, :, i)) for i in axes(x, 2)]
function h(dist, x)
    if all(i -> DistributionsAD.insupport(dist, view(x, :, i)), axes(x, 2))
        logx = log.(x)
        return DistributionsAD._logpdf(dist.normal, logx) - vec(sum(logx; dims = 1))
    else
        return [DistributionsAD._logpdf(dist, view(x, :, i)) for i in axes(x, 2)]
    end
end

@btime f($dist, $x) #   20.266 μs (17 allocations: 29.63 KiB)
@btime g($dist, $x) #   37.591 μs (503 allocations: 68.13 KiB)
@btime h($dist, $x) #   11.583 μs (14 allocations: 21.47 KiB)

@btime f($dist, $y) #   37.841 μs (702 allocations: 73.97 KiB)
@btime g($dist, $y) #   36.997 μs (499 allocations: 67.50 KiB)
@btime h($dist, $y) #   37.237 μs (500 allocations: 67.53 KiB)
```

BTW, benchmarking variant `h2` of `h` that does not precompute `logx = log.(x)`
```julia
@btime h2($dist, $x) #   19.859 μs (15 allocations: 29.41 KiB)
@btime h2($dist, $y) #   37.209 μs (500 allocations: 67.53 KiB)
```
After changing
```julia
function _logpdf(d::TuringMvLogNormal, x::AbstractVector{T}) where {T<:Real}
    return insupport(d, x) ? (logx = log.(x); _logpdf(d.normal, logx) - sum(logx)) : -T(Inf)
end
```
I get
```julia
@btime g($dist, $x) #   26.275 μs (403 allocations: 52.50 KiB)

@btime g($dist, $y) #   26.049 μs (400 allocations: 52.03 KiB)
@btime h($dist, $y) #   26.614 μs (401 allocations: 52.06 KiB)
```

Overall, I think it is more important to improve performance in the case where all samples are in the support of the distribution, since that should be the more common case. IMO the benchmarks suggest implementing alternative `h` and pre-computing `log.(x)` in `_logpdf(d::TuringMvLogNormal, x::AbstractVector{<:Real})`. In a second step one should consider fixing potential type stability problems in `_logpdf(d::TuringMvLogNormal, x::AbstractMatrix{<:Real})`.

mohamed82008: Thanks for this. Pre-computing `log.(x)` is important here. I will fix it thanks.

mohamed82008:
The `MatrixBeta` is fixed here but it requires https://github.com/FluxML/Zygote.jl/pull/477.

src/multivariate.jl
devmotion: The main point I was trying to make is that the best performance can't be achieved by pre-computing `log.(x)` alone. In my benchmarks above, it really requires implementation `h` AND pre-computing `log.(x)`. I can reproduce these findings also with your latest commits (after fixing a newly introduced bug, see below), and implementation `h` still is the fastest one with the fewest allocations (14 instead of18, and 401 instead of 703).

BTW the latest commits introduced a bug in the implementation of the logpdf for vectors. `log.(x)` should be precomputed only if `x` is in the support, the current implementation throws `DomainError`s if `x` is not in the support (see my implementation above).

mohamed82008: I used your implementation thanks :)

src/univariate.jl
```diff
@@ -1,16 +1,50 @@
...
+    return TuringUniform{T}(T(a), T(b))
+end
+Distributions.logpdf(d::TuringUniform, x::Real) = uniformlogpdf(d.a, d.b, x)
+
+Distributions.Uniform(a::TrackedReal, b::Real) = TuringUniform{TrackedReal}(a, b)
+Distributions.Uniform(a::Real, b::TrackedReal) = TuringUniform{TrackedReal}(a, b)
+Distributions.Uniform(a::TrackedReal, b::TrackedReal) = TuringUniform{TrackedReal}(a, b)
+Distributions.logpdf(d::Uniform, x::TrackedReal) = uniformlogpdf(d.a, d.b, x)
+
+uniformlogpdf(a, b, x) = -log(b - a)
```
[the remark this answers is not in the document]
mohamed82008: True.

src/univariate.jl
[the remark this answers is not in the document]
mohamed82008: So in Turing we expect that the gradient is never `nothing` even if the output doesn't depend on the input. This is consistent with `ForwardDiff`. Because of how Tracker and Zygote work, the gradient is sometimes `nothing` if the output of the function is a constant. Perhaps we should consider the `nothing` case in Turing first then we can drop the workaround here. Let's do this in a separate PR.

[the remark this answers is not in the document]
mohamed82008: Hmm actually it's a bit more complicated. In Tracker, constant functions error https://github.com/FluxML/Tracker.jl/issues/48. Hence the need for a workaround here. The Zygote one is just to be consistent but is not necessary if `nothing` is handled properly in Turing.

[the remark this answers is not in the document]
mohamed82008: https://github.com/TuringLang/Turing.jl/issues/1081

mohamed82008:
Once https://github.com/FluxML/Zygote.jl/pull/477 is merged, a new Zygote release is made, and tests pass, this PR can probably be merged.

src/univariate.jl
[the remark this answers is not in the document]
devmotion: Isn't the (main) problem with Tracker that the output types should depend on the input (types), even if the output is constant? E.g., `Tracker.gradient(x -> one(eltype(x)), [0.5])` (just took the example from the Tracker issue that you linked) returns the expected gradient of 0.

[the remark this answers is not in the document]
mohamed82008: Yes this would have to be fixed in Distributions then.

mohamed82008:
If tests pass and there are no objections, I will merge this in a day.

PR #29, opened 2020-02-16
Categorical and Dirichlet

mohamed82008:
This PR defines our versions of `Categorical` and `Dirichlet` distributions.


devmotion:
What exactly are the differences to the implementation of `DiscreteNonParametric` and `Dirichlet` in Distributions?

mohamed82008:
The following examples fail with the versions in Distributions but work with our versions. I don't recall the implementation difference exactly but can check it a bit later.
```julia
using Distributions, Tracker

Tracker.gradient(x -> logpdf(Dirichlet(x), [0.5, 0.5]), ones(2))
Tracker.gradient(x -> sum(logpdf(Categorical(x), [1, 1])), ones(2) ./ 2)
```

mohamed82008:
@devmotion In response to your question there are 2 differences between `Dirichlet` and `TuringDirichlet`:
1. `TuringDirichlet` allows abstract vectors for `alpha`, while Dirichlet only accepts Vector
2. `TuringDirichlet` uses the vectorized sum in the constructor and `logpdf` which are faster than the loop implementations when using `Tracker` and `Zygote` AD

For `Categorical`, I found a way simpler fix, so I removed `TuringDiscreteNonParametric`.

devmotion:
I see! I guess, the first point should be addressed upstream as well at some point.

mohamed82008:
Actually the loop version is likely faster than the vectorized version for non-AD applications so I don't know if the solution should be upstream or not. Anyways, that's a discussion for another day.

devmotion:
I was referring to the first point (AbstractVector), I assume the second one (sum in logpdf and gradients) still has to be addressed in DistributionsAD.

PR #32, opened 2020-02-23
Many more tests and test fixes

mohamed82008:
This PR builds on #19 and #29 and adds many more tests and test fixes. There is no new functionality here.


mohamed82008:
So the tests here are taking too long, Travis is timing out. GitHub actions have a time limit of 6 hours so I am switching. 

There is also a mysterious segfault which happens in different places in different operating systems and doesn't happen when I try to reduce the tests. I am not sure what to do to avoid them but I found that simply re-ordering the tests can avoid the issue. This is definitely an issue upstream in Julia itself but without a MWE, I can't report a bug that takes 1 hour to reproduce. Let's see.

mohamed82008:
I will merge and release based on local tests passing when broken down into pieces. I will come back to fix  CI here later. I know this is not ideal, but if some bug pops up in real code, I promise to track it down as soon as possible :)

PR #33, opened 2020-02-24
Run CI tests only once

devmotion:
Noticed that Travis runs the tests twice for PRs, which should be avoided by this change.


mohamed82008:
I am actually getting rid of Travis altogether in #32 because it times out quickly. I replaced it with GitHub actions which has a longer time limit of 6 hours apparently.

PR #36, opened 2020-02-24
Remove definition of eachcol

devmotion:
Compat already implements `eachcol` for old Julia versions.


mohamed82008:
Thanks @devmotion !

PR #37, opened 2020-02-25
Minor performance and bug fixes (lessons learnt from TuringExamples)

mohamed82008:
This PR does some changes that led to speedups when running the benchmarks in TuringExamples. Some changes are bug fixes as well.


src/common.jl
```diff
@@ -1,13 +1,17 @@
 ## Generic ##
 
-function vcatmapreduce(f, args...)
-    init = vcat(f(first.(args)...,))
-    zipped_args = zip(args...,)
-    return mapreduce(vcat, drop(zipped_args, 1); init = init) do zarg
-        f(zarg...,)
+_istracked(x) = false
+_istracked(x::AbstractArray{<:TrackedReal}) = true
```
torfjelde: Should we also have a 
```julia
_istracked(x::TrackedArray) = true
```
?

mohamed82008: No need. This is only to see if we need to call `vcat` or not to return a `TrackedArray`.

Issue #34, opened 2020-02-24
Tests taking too long and failing


mohamed82008:
Currently, the tests are taking too long for Travis so I switched to GitHub actions in #32. But the tests are still failing with segfaults in different places for different Julia versions. I think it's worth refactoring the tests into multiple stages to reduce time of each. Also we need to identify the causes of segfaults and open issues in Julia's repo.

yebai:
Might be related to Zygote? Since Zygote is the only part that interacts with Julia's compiler and has a history of causing segfaults.

mohamed82008:
Yes I would guess so. I can try commenting out the Zygote tests or allowing them to fail.

yebai:
Or, separate all Zygote tests into a separate stage.

mohamed82008:
So the tests are now separated in stages. Also I think I figured out why Zygote is segfaulting. I think it's cache is overflowing at some point because I differentiate so many functions. Calling `Zygote.refresh()` every function call helps but makes the tests painfully slow. I will experiment locally until I find a sweet frequency of calling `refresh` to get both speed and no-segfaulting.

PR #38, opened 2020-02-28
Call Zygote.refresh every 10 distributions in tests

mohamed82008:
This PR fixes the spontaneous segfaulting that happens in Zygote tests. It worked locally when I called refresh after every distribution but it took a long time. Let's see if this passes tests.


mohamed82008:
Fixes the remaining of #34.

PR #44, opened 2020-03-16
ReverseDiff support

mohamed82008:
This PR adds support for ReverseDiff.


mohamed82008:
I believe this is ready for a review. Tests pass locally. Let's see what CI says.

src/filldist.jl
```diff
@@ -11,7 +11,7 @@ const FillVectorOfUnivariate{
 function filldist(dist::UnivariateDistribution, N::Int)
     return product_distribution(Fill(dist, N))
 end
-filldist(d::Normal, N::Int) = MvNormal(fill(d.μ, N), d.σ)
+filldist(d::Normal, N::Int) = TuringMvNormal(fill(d.μ, N), d.σ)
```
devmotion: Maybe use `Fill` instead of `fill`?

mohamed82008: `Fill` doesn't work well with Tracker. `fill` gives a `TrackedArray`.

devmotion: Ah OK. But isn't the point here that we don't need an actual array?

mohamed82008: TrackedArrays are better for broadcasting. We have broadcasting in MvNormal logpdf.

src/reversediff.jl
```diff
@@ -0,0 +1,235 @@
+const RTR = ReverseDiff.TrackedReal
+const RTV = ReverseDiff.TrackedVector
+const RTM = ReverseDiff.TrackedMatrix
+import SpecialFunctions: logbeta
+import Distributions: Gamma
+
+ReverseDiff.@forward binomlogpdf(n::Int, p::RTR, x::Int) = begin
```
yebai: Maybe add some comments explaining what's happening here.

mohamed82008: Sure, this is using ForwardDiff for the function.

src/filldist.jl
devmotion: I see, thanks for explaining!

PR #52, opened 2020-03-21
fix project.toml

GiggleLiu:
It requires an old Combinatorics package, which may cause conflicts with other packages.


mohamed82008:
Thanks @GiggleLiu!

PR #58, opened 2020-03-28
Custom gradient macro and broadcasting rework for ReverseDiff

mohamed82008:
This PR defines a macro that makes it possible to define custom gradients using a user-friendly syntax. In this PR, I also combine the implementations of broadcasting AD in Tracker and ReverseDiff to make a fast and correct implementation of broadcasting for ReverseDiff.

There are 4 implementations of RD custom adjoints for broadcasting in this PR:
1. A simple fallback on `broadcast` for which custom adjoints are defined in ReverseDiff for the simple unary and binary functions in DiffRules, e.g. +, -, log, exp, etc. The functions all take in numbers and produce a number.
2. A generalization of the custom adjoint for `broadcast` for arbitrary functions. This is generally fast for arrays. Again this works only for inputs that are numbers or arrays of numbers. Broadcastable struct inputs (or Ref) are also supported if not tracked, but not arrays of structs.
3. A port of Tracker's broadcasting implementation for RD. This implementation only works for input numbers and arrays of numbers, or input untracked structs and arrays of structs. Tracked non-numeric inputs give an error with this implementation (and in Tracker). Additionally, broadcasting functions with non-numeric outputs give a silent wrong result in this implementation (and in Tracker) but this is avoided here. When comparing the RD port of Tracker's implementation to Tracker itself, there is a slightly larger overhead to RD's tracking mechanism than Tracker's it seems, but this is small enough that it can be ignored for now. The Tracker port was also observed to be much faster than implementation 2 above when broadcasting over tracked reals. However, when only broadcasting over arrays, implementation 2 was observed to be faster than implementation 3.
4. A fallback that simply outputs an array of `TrackedReal`s. This is sometime the only correct implementation when dealing with structs of tracked fields in the input, or broadcasting a closure of tracked variables.

In this PR, I use:
- Implementation 1 for the functions captured by `DiffRules`,
- Implementation 2 for arbitrary functions (not closures) when no argument is a real number, all array arguments are arrays of real numbers, and the output of the broadcasting can be proven to be an array of real numbers,
- Implementation 3 for arbitrary functions with real number arguments where all array arguments are arrays of real numbers, and the output of the broadcasting can be proven to be an array of real numbers,
- Implementation 4 when broadcasting closures or broadcasting over structs and arrays of structs. By default, any closure, struct or array of structs is assumed to contain tracked variables. However, if the closure, struct or array of structs arguments is wrapped with `NotTracked`, either implementation 2 or 3 will be used instead. These are more performant because they output a `TrackedArray` instead of the `Array{<:TrackedReal}` which implementation 4 outputs. We can use `NotTracked` for example to use the more efficient broadcasting for `logpdf` of `filldist(::Bernoulli{Float64})` and `arraydist(::Bernoulli{Float64})`.


src/reversediffx.jl
```diff
@@ -0,0 +1,582 @@
...
+    end
+The `@grad` macro provides a way for the users to define custom adjoints for single-output functions wrt to their input numbers or arrays.
+"""
+macro grad(expr)
+    if @capture(expr, 
+        (f_(xs__) where {T__} = body_) | 
+        (f_(xs__) = body_) | 
+        (function f_(xs__) body_ end) | 
+        (function f_(xs__) where {T__} body_ end)
+    )
```
devmotion: Wouldn't it be easier to directly use `MacroTools.splitdef`?

devmotion: I guess that would allow to remove also other things such as `getargs_expr`.

src/reversediffx.jl
```diff
@@ -0,0 +1,582 @@
...
+    f::F
+end
+(f::NotTracked{<:Union{Function, Type}})(args...; kwargs...) = f.f(args...; kwargs...)
+
+istypeorclosure(::F) where {F} = _istypeorclosure(F)
+istypeorclosure(::AbstractArray{F}) where {F} = _istypeorclosure(F)
+istypeorclosure(::Base.RefValue{F}) where {F} = _istypeorclosure(F)
+istypeorclosure(::AbstractArray{<:Real}) = false
+istypeorclosure(::Real) = false
+@generated _istypeorclosure(::Type{F}) where {F} = :($(fieldcount(F) > 0))
```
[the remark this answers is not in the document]
devmotion: BTW according to the docstring `fieldcount` throws an error if the type is too abstract.

src/reversediffx.jl
```diff
@@ -0,0 +1,582 @@
...
+
+@inline mayhavetracked(b) = istypeorclosure(b)
+@inline mayhavetracked(b::NotTracked) = false
+@inline mayhavetracked(b::Base.RefValue{<:NotTracked}) = false
+@inline mayhavetracked(b::Broadcasted) = mayhavetracked(b.f) || any(mayhavetracked, b.args)
```
devmotion: Are the `@inline` statements needed here?

src/reversediffx.jl
```diff
@@ -0,0 +1,582 @@
...
+end
+
+@generated function onlyrealarrays(args::T) where {T <: Tuple}
+    o = all(map(x -> (x <: AbstractArray{<:Real} || !(x <: AbstractArray)), T.types))
+    return :($o)
+end
+@generated function anyreals(args::T) where {T <: Tuple}
+    o = any(map(x -> x <: Real, T.types))
+    return :($o)
+end
```
devmotion: Maybe also just use regular functions here?
```julia
anyreals(args::Tuple) = tuple_type_anyreals(typeof(args))

tuple_type_anyreals(::Type{Tuple{}}) = false
function tuple_type_anyreals(::Type{T}) where T<:Tuple
	return Base.tuple_type_head(T) <: Real || tuple_type_anyreals(Base.tuple_type_tail(T))
end
```
With Julia >= 1.1, one could also use
```julia
anyreals(args::Tuple) = any(x <: Real for x in fieldtypes(typeof(args)))
```
but the former seems to be easier for the compiler.

src/reversediffx.jl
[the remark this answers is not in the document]
phipsgabler: Also: the argument of `fieldcount` is `@nospecialize`d (at least in 1.3).  Probably the implementors reasoned that specialization in this case is worse than the repeated calculation?  I'm not very well versed in these kinds of optimizations, though...

src/reversediffx.jl
```diff
@@ -0,0 +1,582 @@
...
+remove_not_tracked(f::Base.RefValue{<:NotTracked}) = Ref(remove_not_tracked(f[]))
+remove_not_tracked(f::Base.RefValue{<:NotTracked{<:AbstractArray}}) = remove_not_tracked(f[])
+function remove_not_tracked(b::Broadcasted{style}) where {style}
+    return Broadcasted{style}(remove_not_tracked(b.f), remove_not_tracked.(b.args), b.axes)
+end
+
+@generated function onlyrealarrays(args::T) where {T <: Tuple}
+    o = all(map(x -> (x <: AbstractArray{<:Real} || !(x <: AbstractArray)), T.types))
+    return :($o)
+end
```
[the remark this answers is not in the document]
mohamed82008: This is not a correct replacement. If there is a struct that is not real or an array, it should return true.

src/reversediffx.jl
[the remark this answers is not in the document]
mohamed82008: The generated function is needed here.
```julia
julia> f(::Type{T}) where {T} = fieldcount(T) > 0

julia> @generated g(::Type{T}) where {T} = :($(fieldcount(T) > 0))
g (generic function with 1 method)

julia> @code_llvm g(Int)

;  @ REPL[10]:1 within `g'
; Function Attrs: uwtable
define i8 @julia_g_17479(%jl_value_t addrspace(10)*) #0 {
top:
; ┌ @ REPL[10]:1 within `macro expansion'
   ret i8 0
; └
}

julia> @code_llvm f(Int)

;  @ REPL[9]:1 within `f'
; Function Attrs: uwtable
define i8 @julia_f_17480(%jl_value_t addrspace(10)*) #0 {
top:
  %gcframe = alloca %jl_value_t addrspace(10)*, i32 3, align 16
  %1 = bitcast %jl_value_t addrspace(10)** %gcframe to i8*
  call void @llvm.memset.p0i8.i32(i8* align 16 %1, i8 0, i32 24, i1 false)
  %2 = call %jl_value_t*** inttoptr (i64 1720669888 to %jl_value_t*** ()*)() #5
; ┌ @ reflection.jl:702 within `fieldcount'
; │┌ @ Base.jl:28 within `getproperty'
    %3 = getelementptr %jl_value_t addrspace(10)*, %jl_value_t addrspace(10)** %gcframe, i32 0
    %4 = bitcast %jl_value_t addrspace(10)** %3 to i64*
    store i64 4, i64* %4
    %5 = getelementptr %jl_value_t**, %jl_value_t*** %2, i32 0
    %6 = getelementptr %jl_value_t addrspace(10)*, %jl_value_t addrspace(10)** %gcframe, i32 1
    %7 = bitcast %jl_value_t addrspace(10)** %6 to %jl_value_t***
    %8 = load %jl_value_t**, %jl_value_t*** %5
    store %jl_value_t** %8, %jl_value_t*** %7
    %9 = bitcast %jl_value_t*** %5 to %jl_value_t addrspace(10)***
    store %jl_value_t addrspace(10)** %gcframe, %jl_value_t addrspace(10)*** %9
    %10 = load i8, i8* inttoptr (i64 156646900 to i8*), align 4
; │└
; │ @ reflection.jl:704 within `fieldcount'
   %11 = and i8 %10, 1
   %12 = icmp eq i8 %11, 0
   br i1 %12, label %L12, label %L9

L9:                                               ; preds = %top
; │ @ reflection.jl:705 within `fieldcount'
; │┌ @ boot.jl:288 within `ArgumentError'
    %13 = bitcast %jl_value_t*** %2 to i8*
    %14 = call noalias nonnull %jl_value_t addrspace(10)* @jl_gc_pool_alloc(i8* %13, i32 1400, i32 16) #2
    %15 = bitcast %jl_value_t addrspace(10)* %14 to %jl_value_t addrspace(10)* addrspace(10)*
    %16 = getelementptr %jl_value_t addrspace(10)*, %jl_value_t addrspace(10)* addrspace(10)* %15, i64 -1
    store %jl_value_t addrspace(10)* addrspacecast (%jl_value_t* inttoptr (i64 158151344 to %jl_value_t*) to %jl_value_t addrspace(10)*), %jl_value_t addrspace(10)* addrspace(10)* %16
    %17 = bitcast %jl_value_t addrspace(10)* %14 to %jl_value_t addrspace(10)* addrspace(10)*
    store %jl_value_t addrspace(10)* addrspacecast (%jl_value_t* inttoptr (i64 160740832 to %jl_value_t*) to %jl_value_t addrspace(10)*), %jl_value_t addrspace(10)* addrspace(10)* %17, align 8
; │└
   %18 = addrspacecast %jl_value_t addrspace(10)* %14 to %jl_value_t addrspace(12)*
   %19 = getelementptr %jl_value_t addrspace(10)*, %jl_value_t addrspace(10)** %gcframe, i32 2
   store %jl_value_t addrspace(10)* %14, %jl_value_t addrspace(10)** %19
   call void @jl_throw(%jl_value_t addrspace(12)* %18)
   unreachable

L12:                                              ; preds = %top
; │ @ reflection.jl:708 within `fieldcount'
; │┌ @ Base.jl:28 within `getproperty'
    %20 = load %jl_value_t addrspace(10)*, %jl_value_t addrspace(10)** inttoptr (i64 156646856 to %jl_value_t addrspace(10)**), align 8
    %21 = icmp eq %jl_value_t addrspace(10)* %20, null
    br i1 %21, label %fail, label %pass

fail:                                             ; preds = %L12
    call void @jl_throw(%jl_value_t addrspace(12)* addrspacecast (%jl_value_t* inttoptr (i64 159065248 to %jl_value_t*) to %jl_value_t addrspace(12)*))
    unreachable

pass:                                             ; preds = %L12
    %22 = getelementptr %jl_value_t addrspace(10)*, %jl_value_t addrspace(10)** %gcframe, i32 1
    %23 = load %jl_value_t addrspace(10)*, %jl_value_t addrspace(10)** %22
    %24 = getelementptr %jl_value_t**, %jl_value_t*** %2, i32 0
    %25 = bitcast %jl_value_t*** %24 to %jl_value_t addrspace(10)**
    store %jl_value_t addrspace(10)* %23, %jl_value_t addrspace(10)** %25
; └└
  ret i8 0
}
```

src/reversediffx.jl
```diff
@@ -0,0 +1,582 @@
...
+const RTM = ReverseDiff.TrackedMatrix
+const RTA = ReverseDiff.TrackedArray
+const RDBroadcasted{F, T} = Broadcasted{<:Any, <:Any, F, T}
+
+"""
+    f(x) = dot(x, x)
+    f(x::ReverseDiff.TrackedVector) = ReverseDiff.track(f, x)
+    ReverseDiff.@grad function f(x)
+        xv = ReverseDiff.value(x)
+        return dot(xv, xv), ∇ -> (∇ * 2 * xv,)
```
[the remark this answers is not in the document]
mohamed82008: Good catch, thanks!

src/reversediffx.jl
mohamed82008: Possibly, I will try.

src/reversediffx.jl
mohamed82008: Let me see.

src/reversediffx.jl
mohamed82008: Fair enough, I used a generated function because it is a sure fire way of doing this at compile time. Sometimes large tuples and recursion cause inference issues but not this time it seems.

src/reversediffx.jl
```diff
@@ -0,0 +1,582 @@
...
+    function back(Δ)
+        start = ntuple(i -> 0, Val(ndims(Δ)))
+        Δs = [begin
+          dim_xs = 1:ndims(xs)
+          till_xs = ntuple((i -> i in dims ? (i in dim_xs ? size(xs,i) : 1) : 0), Val(ndims(Δ)))
+          xs_in_Δ = ntuple(i -> till_xs[i] > 0 ? (start[i]+1:start[i]+till_xs[i]) : Colon(), Val(ndims(Δ)))
+          d = reshape(Δ[xs_in_Δ...],size(xs))
+          start = start .+ till_xs
+          d
+        end for xs in Xs]
```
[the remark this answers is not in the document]
mohamed82008: 👍

src/reversediffx.jl
mohamed82008: Seems not.

src/reversediffx.jl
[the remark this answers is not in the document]
mohamed82008: The following is correct but longer and depends on the compiler inlining all the calls at compile time to get a similar performance to the generated function.
```julia
onlyrealarrays(args::Tuple) = onlyrealarray(first(args)) && onlyrealarrays(Base.tail(args))
onlyrealarrays(args::Tuple{}) = true
onlyrealarray(::AbstractArray{<:Real}) = true
onlyrealarray(::AbstractArray) = false
onlyrealarray(::Any) = true
```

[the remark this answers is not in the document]
mohamed82008: Replaced.

src/reversediffx.jl
mohamed82008: Done.

src/reversediffx.jl
mohamed82008: Done.

mohamed82008:
This PR is almost ready but I want to give kwargs a try because @AStupidBear seems to need them. `MacroTools.splitdef` will make this easier. Another thing we may want to consider in Turing/DPPL is to make `x .~ Normal.(m)` syntactic sugar for:
```julia
d = arraydist(LazyArrays.@~(Normal.(m)))
vi.logp[] += logpdf(d, x)
```
This will then fuse the broadcasting making it return a correct `TrackedArray` in both `Tracker` and `ReverseDiff`. For example:
```julia
using Distributions, LazyArrays, ReverseDiff, DistributionsAD

m = ReverseDiff.track(rand(2))
d = LazyArrays.@~(Normal.(m))
logpdf.(d, rand(2))
# Equivalent to logpdf.(Normal.(m), rand(2))
```
returns a `TrackedArray`, but:
```julia
d = Normal.(m)
logpdf.(d, rand(2))
```
returns an array of tracked reals in RD and a wrong result (a normal array) in Tracker.

devmotion:
> Another thing we may want to consider in Turing/DPPL is to make x .~ Normal.(m) syntactic sugar for:
> d = arraydist(LazyArrays.@~(Normal.(m)))
> vi.logp[] += logpdf(d, x)

Wouldn't it probably be more efficient if `x .~ Normal.(m)` would be syntactic sugar for `x ~ MvNormal(m)` if `m isa AbstractVector`? I guess, ideally we would not always use LazyArrays but exploit optimized multivariate and matrixvariate distributions if possible.

BTW what would be the difference then between `x .~ Normal.(m)` and `x ~ Normal.(m)`? Should the first expression be used if `x` already exists (e.g., `x = Matrix{T}(undef, n, m)` before) similar to `x .= m` in standard broadcasting?

mohamed82008:
>Wouldn't it probably be more efficient if x .~ Normal.(m) would be syntactic sugar for x ~ MvNormal(m) if m isa AbstractVector?

Yes I was just using `Normal` as a convenient example, replace it with `Bernoulli` for example. But even for `Normal` it may be faster this way.

> Should the first expression be used if x already exists (e.g., x = Matrix{T}(undef, n, m) before) similar to x .= m in standard broadcasting?

That's similar to the current convention. Perhaps we can make all broadcasting on the rhs of `~` or `.~` lazy then. That would make `x ~ Normal.(m)` fast. Let's discuss this further in another issue/PR.

Project.toml
```diff
@@ -10,11 +10,13 @@ Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
 FillArrays = "1a297f60-69ca-5386-bcde-b61e274b549b"
 ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+MacroTools = "1914dd2f-81c6-5fcd-8719-6d5c9610ff09"
```
devmotion: If only `splitdef` is needed from MacroTools, I suggest using [ExprTools.jl](https://github.com/invenia/ExprTools.jl) instead. It's a more light-weight package without additional dependencies that contains `splitdef` and `combinedef`.

mohamed82008: I also use `@capture`.

ChrisRackauckas:
@mohamed82008 @AStupidBear I'd be willing to accept these overloads to DiffEqBase, just like the others:

https://github.com/SciML/DiffEqBase.jl/blob/master/src/solve.jl#L200-L234

PR #59, opened 2020-03-28
Update and fix CI tests

devmotion:
This PR updates all CI tests to run on Julia 1.4 and 1.0 and reduces the frequency of TagBot and CompatHelper to once per day. Although intended otherwise, CI tests were not run on Julia 1.0 before (`[1.0, 1.3]` is interpreted as `[1, 1.3]`).

### DIFF
--- a/.github/workflows/CompatHelper.yml
+++ b/.github/workflows/CompatHelper.yml
@@ -1,26 +1,19 @@
 name: CompatHelper
 
 on:
-  # push:
   schedule:
-    - cron: '00 * * * *'
+    - cron: '00 00 * * *'
 
 jobs:
-  build:
-    runs-on: ${{ matrix.os }}
-    strategy:
-      matrix:
-        julia-version: [1.3]
-        julia-arch: [x86]
-        os: [ubuntu-latest]
+  CompatHelper:
+    runs-on: ubuntu-latest
     steps:
       - uses: julia-actions/setup-julia@latest
         with:
-          version: ${{ matrix.julia-version }}
-      - name: Install dependencies
+          version: 1.4
+      - name: Pkg.add("CompatHelper")
         run: julia -e 'using Pkg; Pkg.add("CompatHelper")'
-      - name: CompatHelper.main
+      - name: CompatHelper.main()
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          JULIA_DEBUG: CompatHelper
         run: julia -e 'using CompatHelper; CompatHelper.main()'
--- a/.github/workflows/ForwardDiff_Tracker.yml
+++ b/.github/workflows/ForwardDiff_Tracker.yml
@@ -5,25 +5,31 @@ on:
     branches:
       - master
   pull_request:
-    types: [opened, synchronize, reopened]
 
 jobs:
   test:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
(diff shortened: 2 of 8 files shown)


devmotion:
ReverseDiff tests fail on Julia 1.0: https://github.com/TuringLang/DistributionsAD.jl/pull/59/checks?check_run_id=541626326
Since the tests have never been run on Julia 1.0, I don't know if this is known.

devmotion:
Zygote and ForwardDiff/Tracker tests were aborted due to OutOfMemory errors.

mohamed82008:
Thanks for the PR!

> TagBot and CompatHelper to once per day

Why is this necessary?

devmotion:
Of course, it is not necessary but it's a lot of useless compute power needed for running these tests (24 times more exactly :smile:). I've seen some discussions on Slack about this waste of resources in the Julia package ecosystem and that's why, e.g., once per day is now even suggested in [CompatHelper's installation instructions](https://github.com/bcbi/CompatHelper.jl#installation). I guess usually both tasks are not very urgent and additionally if you know of an updated dependency you can always submit a manual PR.

mohamed82008:
I am ok with CompatHelper getting run once a day but TagBot is important to run as fast as possible if we need to make hot fix releases, e.g. fixing a bug introduced in a previous release made on the same day.

devmotion:
> I am ok with CompatHelper getting run once a day but TagBot is important to run as fast as possible if we need to make hot fix releases, e.g. fixing a bug introduced in a previous release made on the same day.

I don't see why TagBot is needed for this? TagBot is not required for Julia releases, it just ensures that the git history is "nice" by adding the corresponding tags. The important thing is that the new version is listed in the registry which does not depend on TagBot.

mohamed82008:
>The important thing is that the new version is listed in the registry which does not depend on TagBot.

I think users cannot add the latest version until the tag is made, no?

devmotion:
Sure they can. As I said, TagBot is completely optionally and just creates some nice git history. The releases are available as soon as the PR to the registry is merged. Packages are not registered based on tags but on the git hash: https://github.com/JuliaRegistries/General/blob/master/D/DistributionsAD/Versions.toml

mohamed82008:
The tests seem to fail on 32 bit Julia 1.0 due to some peculiar LLVM out of memory errors. Perhaps we can drop those tests?

mohamed82008:
>Sure they can. As I said, TagBot is completely optionally and just creates some nice git history. The releases are available as soon as the PR to the registry is merged. Packages are not registered based on tags but on the git hash: https://github.com/JuliaRegistries/General/blob/master/D/DistributionsAD/Versions.toml

Fair enough. Thanks for explaining.

devmotion:
> The tests seem to fail on 32 bit Julia 1.0 due to some peculiar LLVM out of memory errors. Perhaps we can drop those tests?

They fail on Julia 1.4. But sure, we could drop 32bit tests here if you think it's mainly a problem with running too many tests and/or we don't want to support 32bit. I'm not sure how many people actually use 32bit systems these days :man_shrugging:

mohamed82008:
Ya removing 32 bit tests seems reasonable. The LLVM failure seems like a Julia/LLVM issue anyways so there isn't much we can do about it here. Maybe just open an issue in the julia repo?

mohamed82008:
Thanks @devmotion!

PR #63, opened 2020-04-10
Refactor

mohamed82008:
This PR refactors some of the code in the package for performance. In particular:

1. Less emphasis is put on returning a `TrackedArray` when it is difficult to. Previously, I went to extreme lengths `vcat`ing every array of `TrackedReal` to a `TrackedArray`. However, for large arrays this comes at a significant performance cost and is known to cause a `StackOverFlow` error for Tracker. It is also not that hard for the user to do `reduce(vcat, array)` if they would like a `TrackedArray` explicitly.

2. `arraydist(::AbstractVector{<:Normal})` now doesn't return `MvNormal`. Converting to `MvNormal` involves concatenation which again I am trying to avoid. 

3. I added glue code to support efficient lazy broadcasting using `arraydist(::BroadcastArray)`. When `LazyArrays.jl` is loaded, the function `lazyarray` will be defined as `lazyarray(f, x...) = LazyArray(Base.broadcasted(f, x...))` and exported. This can be used in `d = arraydist(lazyarray(Normal, ::TrackedArray))` instead of `d = arraydist(Normal.(::TrackedArray))`. The former will do the right thing when calling `logpdf.(d, x)` in both `Tracker` and `ReverseDiff` returning a `TrackedArray`, while the latter returns a wrong result in `Tracker` and returns an array of `TrackedReal` in `ReverseDiff`.

4. This PR ports all the Tracker type piracy fixes from Bijectors to have only a single pirate.

5. This PR removes the ReverseDiff `@grad` macro from DistributionsAD as it will be soon merged in ReverseDiff (https://github.com/JuliaDiff/ReverseDiff.jl/pull/123). So this PR will only pass tests after the ReverseDiff PR gets merged and released.

6. This PR makes ReverseDiff an optional dependency to reduce the loading time (CC: @torfjelde).


torfjelde:
A bit uncertain whether or not dropping `vcat` is the best idea atm :confused: Ref: https://github.com/TuringLang/Bijectors.jl/pull/99#issuecomment-612398170

mohamed82008:
I got rid of `mapvcat` altogether here. This should make the example in https://github.com/TuringLang/Bijectors.jl/pull/99 work correctly.

torfjelde:
> I got rid of mapvcat altogether here. This should make the example in TuringLang/Bijectors.jl#99 work correctly.

But this will then be an issue: https://github.com/TuringLang/Bijectors.jl/pull/99#issuecomment-612421429

mohamed82008:
I think we just need to be consistent in TuringLang. We can't control what the user writes. If only produce TrackedArrays and the user adds it to an array of TrackedReals, it is not our fault. We need to document this Tracker bug though.

torfjelde:
> If we only produce TrackedArrays and the user adds it to an array of TrackedReals, it is not our fault.

But now we only produce TrackedReals, right? Which is kind of going against what everyone else _attempts_ to do when using Tracker.

Issue #65, opened 2020-04-25
Complex dependencies


marcoct:
This package is great work and the Gen team would love to collaborate and contribute to this. One concern I have is that DistributionsAD brings in some a list of dependencies beyond those of Zygote, Tracker, ForwardDiff, and ReverseDiff,  that includes some potentially unnecessary packages and binaries:

  [861a8166] + Combinatorics v0.7.0
  [ced4e74d] + DistributionsAD v0.1.4
  [e2ba6199] + ExprTools v0.1.1
  [8f5d6c58] + EzXML v1.1.0
  [d8418881] + Intervals v1.1.0
  [94ce4f54] + Libiconv_jll v1.16.0+2
  [78c3b35d] + Mocking v0.7.1
  [f27b6e38] + Polynomials v0.8.0
  [3cdcf5f2] + RecipesBase v1.0.1
  [f269a46b] + TimeZones v1.1.1
  [02c8fc9c] + XML2_jll v2.9.9+4
  [83775a58] + Zlib_jll v1.2.11+9

This chain of dependencies seems to be responsible for many of these:

Combinatorics => Polynomials => Intervals => TimeZones => EzXML

It would be possible to break the chain and remove dependence on Intervals in either DistributionsAD itself, or in Combinatorics, or Polynomials. In any of these cases, if there is a single function that is used from the package one-level down the chain, it might make sense to pay the cost of re-implementing it, instead of adding these dependencies.

mohamed82008:
Hi! I am glad to hear the Gen team is willing to collaborate on this package! I also don't like the current state of DistributionsAD's dependencies. It was mostly driven by a desperation to make things work without thinking too much about loading time and dependencies. I am open to improvement suggestions. As for Combinatorics, we only use the `combinations` function so re-implementing it shouldn't be too hard.

mohamed82008:
Actually Combinatorics can be removed entirely in the next release after https://github.com/JuliaDiff/ReverseDiff.jl/pull/125 is merged and released.

yebai:
thank @marcoct, it would be great to work together on the common set of libraries underlying Turing and Gen, and avoid duplicate efforts where possible. Also, we're happy to move `DistributionsAD` to a more suitable org account in the future, e.g. `JuliaStats` or `JuliaPPL`.

devmotion:
[...]

PR #66, opened 2020-04-26
Drop Zygote dependency and reformat code

devmotion:
This PR removes the Zygote dependency, as discussed in https://github.com/TuringLang/DistributionsAD.jl/issues/65#issuecomment-619516746. Moreover, currently sometimes the explicit `ZygoteRules.@adjoint` and `ZygoteRules.pullback` was used and sometimes only `@adjoint` and `pullback`. I reformatted the code (adhering to the 92 characters limit) so that now consistently the more explict `ZygoteRules.@adjoint` and `ZygoteRules.pullback` are used to show clearly that these implementations use Zygote.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -20,7 +20,6 @@ StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 StatsFuns = "4c63d2b9-4356-54db-8cca-17b64c39e42c"
 Tracker = "9f7883ad-71c0-57eb-9f7f-b5c9e6d3789c"
-Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f"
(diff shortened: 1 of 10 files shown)


devmotion:
BTW maybe we should stop testing Zygote on Julia 1.0 - tests run for almost 4 hours now whereas tests on Julia 1.4 were finished after a bit less than 2 hours. Zygote's README also states explicitly:
> Zygote supports Julia 1.0 onwards, but we highly recommend using Julia 1.3 or later.

mohamed82008:
>BTW maybe we should stop testing Zygote on Julia 1.0 - tests run for almost 4 hours now whereas tests on Julia 1.4 were finished after a bit less than 2 hours.

Happy to stop testing Zygote on Julia 1.0.

devmotion:
I removed the Zygote tests on Julia 1.0, unfortunately it seems Github actions don't respect `[skip ci]` in the commit message...

Issue #68, opened 2020-04-27
rand is incorrect for FillVectorOfUnivariate is incorrect


torfjelde:
`rand` fills the entire vector with the same sample:
```julia
julia> using DistributionsAD
[ Info: Precompiling DistributionsAD [ced4e74d-a319-5a8a-b0ac-84af2272839c]

julia> dist = filldist(DistributionsAD.Gamma(1., 1.), 2)
Distributions.Product{Distributions.Continuous,Distributions.Gamma{Float64},FillArrays.Fill{Distributions.Gamma{Float64},1,Tuple{Base.OneTo{Int64}}}}(v=Distributions.Gamma{Float64}[Distributions.Gamma{Float64}(α=1.0, θ=1.0), Distributions.Gamma{Float64}(α=1.0, θ=1.0)])

julia> rand(dist)
2-element Array{Float64,1}:
 0.4873494685298349
 0.4873494685298349

julia> rand(dist)
2-element Array{Float64,1}:
 1.794513850833854
 1.794513850833854
```

mohamed82008:
It seems that this is a Distributions.jl bug. We use `Product` to implement `VectorOfUnivariate` and `FillVectorOfUnivariate`.

mohamed82008:
Closed via #69.

PR #69, opened 2020-04-27
Fix for #68

torfjelde:
Fix for #68

TODO: add test

### DIFF
--- a/src/filldist.jl
+++ b/src/filldist.jl
@@ -65,6 +65,10 @@ function _flat_logpdf_mat(dist, x)
     end
 end
 
+function Distributions.rand(rng::Random.AbstractRNG, d::FillVectorOfUnivariate)
+    return rand(rng, d.v.value, length(d))
+end
+
 const FillMatrixOfUnivariate{
     S <: ValueSupport,
     T <: UnivariateDistribution{S},


src/filldist.jl
```diff
@@ -65,6 +65,10 @@ function _flat_logpdf_mat(dist, x)
     end
 end
 
+function Distributions.rand(rng::Random.AbstractRNG, d::FillVectorOfUnivariate)
```
mohamed82008, line 68: This should be covered by `Distributions.Product`.

mohamed82008: But of course it is not. So happy to merge this.

torfjelde: Yeah, so last night I didn't get have the time to really look into it, but it seems like it's actually a bug with how `Fill` is handled by `broadcast!`:
```julia
dist = dist = filldist(Gamma(), 2)
x = zeros(2)
broadcast!(dn -> rand(dn), x, dist.v)
```
results in the same issue. Going down a rabit-hole, I think the issue is how `materialize!` (https://github.com/JuliaLang/julia/blob/0bab06f8a7f70d4518ad30e19a4c0a1968c875ad/base/broadcast.jl#L823) acts, which means there's something FillArrays.jl's broadcasting style I think. I'll open an issue over at FillArrays.jl

torfjelde: Seems like good ole' @willtebbutt as been down this road before: https://github.com/JuliaArrays/FillArrays.jl/issues/29#issue-359096800

PR #72, opened 2020-05-04
[WIP] Add eps to probs in discrete distributions

mohamed82008:
This PR adds eps to the probabilities in MvCategorical to avoid `-Inf` from showing when starting from a bad initial value in Turing. The same needs to be done or all the other discrete distributions.

### DIFF
--- a/src/mvcategorical.jl
+++ b/src/mvcategorical.jl
@@ -156,15 +156,16 @@ function Distributions.logpdf(d::MvCategorical{T}, x::AbstractVector{<:Integer})
         return zero(eltype(ps))
     end
 end
-_mv_categorical_logpdf(ps, x) = sum(log, view(ps, x, :))
+_mv_categorical_logpdf(ps, x) = sum(x -> log(x + eps(x)), view(ps, x, :))
 _mv_categorical_logpdf(ps::Tracker.TrackedMatrix, x) = Tracker.track(_mv_categorical_logpdf, ps, x)
 Tracker.@grad function _mv_categorical_logpdf(ps, x)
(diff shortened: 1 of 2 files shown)


mohamed82008:
I think I should make `TuringXXX` constructors for these distributions to keep the normal constructors when they are needed. Maybe the user wants to get `-Inf` and reject all samples with a certain parameter.

devmotion:
Why not work with a categorical distribution that is parameterized by the logit of its probabilities (or some other transformation to the real line) if you run into this problem? Couldn't one then avoid the issue in a numerically better way (i.e., even work with probabilities that are in `(0, eps)`)?

mohamed82008:
>Why not work with a categorical distribution that is parameterized by the logit of its probabilities (or some other transformation to the real line) if you run into this problem? Couldn't one then avoid the issue in a numerically better way (i.e., even work with probabilities that are in (0, eps))?

This is possible of course, and I am starting to doubt whether this feature is actually useful or not. But it might be useful to provide something like out of the box. Note that `logistic` can still give 0 and 1 for some inputs because of underflow so it doesn't completely help with the problem here. Admittedly the problem I am trying to work around here is one that the user should be able to figure out and solve on their own but it's about making their lives easier.

devmotion:
Sure, I guess it's probably still worth including and you can of course end up with 0 or 1 if you compute the probabilities from the transformed parameters. But if you're (only) interested in optimization or estimation, that step would probably be needed only in the very end (if it's needed at all). In any case, my intuition was just that you could perform the gradient computation and optimization steps in R without having to think about what probability value it corresponds to, which I guess could help here.

PR #75, opened 2020-05-06
Define custom adjoint for simplex_logpdf

mohamed82008:
This PR fixes a performance regression in the LDA and hmm_semisup benchmarks in TuringExamples when using the latest release of DistributionsAD compared to the branch used in TuringExamples#master.


mohamed82008:
So while this should make things faster, it doesn't seem to give the same speed that we had in TuringExamples#master. I will keep looking.

Issue #55, opened 2020-03-21
MixtureModels


theogf:
The mixturemodel of distributions.jl has some type assert making it incompatible with most AD.

sivapvarma:
I  was thinking of taking this up as I need it. I am familiar with Zygote and ReverseDiff. I haven't used Tracker before. 

Will this require changes in Distributions.jl or is it enough to add methods here.

mohamed82008:
Hi! Once you get it working, we can decide which code can go in here and which may need to go in Distributions.jl. We may also host some code here temporarily until it is merged and released in Distributions.jl.

PR #79, opened 2020-05-11
[WIP] Mixture Models

sivapvarma:
WIP to fix issue #55. 

Right now I am using Zygote.Buffer which adds Zygote as a dependency, which is probably not desirable. I am not sure how to use Zygote.Buffer without adding Zygote dep.

Any comments are welcome.


src/mixturemodels.jl
```diff
@@ -0,0 +1,46 @@
...
+    #  such that the argument of exp is in a reasonable range
+    #
+
+    K = ncomponents(d)
+    p = probs(d)
+    # use Buffer to avoid mutating arrays.
+    # lp = Vector{eltype(p)}(undef, K)
+    lp = Zygote.Buffer(p, K)
+    m = -Inf   # m <- the maximum of log(p(cs[i], x)) + log(pri[i])
+    @inbounds for i in eachindex(p)
```
mohamed82008, line 19: Try defining a non-mutating function that takes in `p` as input and returns `lp` as output. Use a `map` block. `m` is just `maximum(lp)`.

mohamed82008: For example,
```julia
map(eachindex(p)) do i
   ...
end
```

mohamed82008:
Thanks for the PR @sivapvarma! Please implement the above function in a non-mutating way, it's possible. Then add a `MixtureModel` to the tests.

src/mixturemodels.jl
sivapvarma: jus to be clear, you are fine with using Zygote.Buffer, which means we need to add Zygote to Project.toml.

Your comments are only about refactoring the for loop right ?

devmotion:
Thanks for the PR but IMO this should really be fixed in Distributions. The implementation there (which is mainly copied here) should just use `map` and `logsumexp`, and then probably it works with AD. I think we don't want to use `Zygote.Buffer` and depend on Zygote in DistributionsAD.

sivapvarma:
@devmotion @mohamed82008 I see what you mean now, this can be done without Zygote.Buffer in a non-mutating way. Let me do that and remove Zygote dependency, then we can decide if this should go in Distributions or not.

sivapvarma:
Done. No Zygote dependency anymore. But when I precompile the package, I get warnings about method overwritten. What is the solution to this ?

devmotion:
> But when I precompile the package, I get warnings about method overwritten. What is the solution to this ?

This is expected since you overwrite the definitions in Distributions. IMO the correct solution would be to fix the logpdf implementation in Distributions instead of adding a second one in DistributionsAD.

src/mixturemodels.jl
```diff
@@ -0,0 +1,24 @@
+function _mixlogpdf1(d::AbstractMixtureModel, x)
+    # using the formula below for numerical stability
+    #
+    # logpdf(d, x) = log(sum_i pri[i] * pdf(cs[i], x))
+    #              = log(sum_i pri[i] * exp(logpdf(cs[i], x)))
+    #              = log(sum_i exp(logpri[i] + logpdf(cs[i], x)))
+
```
devmotion, line 7: The comment seems a bit useless now that `logsumexp` is used instead of the manual implementation of the logsumexp trick.
```suggestion
```

src/mixturemodels.jl
```diff
@@ -0,0 +1,24 @@
...
+    K = ncomponents(d)
```
devmotion, line 8: `K` is not used anywhere.
```suggestion
```

src/mixturemodels.jl
```diff
@@ -0,0 +1,24 @@
...
+    pri = probs(d)
+
+    lp = map(eachindex(pri)) do i
+            if pri[i] > 0.0
+                return logpdf(component(d, i), x) + log(pri[i])
+            end
+        end
```
devmotion, line 15: This won't work correctly since `lp[i]` will be `nothing` if `pri[i] <= 0.0`. Maybe better use something like
```suggestion
    indices = findall(!iszero, pri)
    lp = map(indices) do i
        return logpdf(component(d, i), x) + log(pri[i])
    end
```

mohamed82008:
@sivapvarma we still need tests. I agree with @devmotion that if this is a strictly better implementation or at least as good as the one in Distributions.jl even without AD, then it needs to go there.

src/mixturemodels.jl
sivapvarma: I originally wanted to do a list comprehension like
```julia
lp = [ logpdf(component(d, i), x) + log(pri[i]) for i in eachindex(pri) if pri[i] > 0.0 ]
```
which does not have the `nothing`'s in `lp`.  This some how does not work with Zygote. 

Then I tried to do the samething with `map` and assumed the `nothing`'s would be filtered out. I should have tested it out in a REPL first. Thanks for catching this.

devmotion:
Will be fixed upstream by https://github.com/JuliaStats/Distributions.jl/pull/1308.

devmotion:
The upstream PR is merged and part of Distributions 0.25.0.

PR #84, opened 2020-06-06
Fix Wishart and InverseWishart

devmotion:


src/matrixvariate.jl
```diff
@@ -40,15 +40,15 @@ function TuringWishart(df::T, S::AbstractMatrix) where {T <: Real}
...
+    R = Base.promote_eltype(T, logc0)
+    return TuringWishart(R(df), C, R(logc0))
 end
 
-function _wishart_c0(df::Real, C::Cholesky)
+function _wishart_logc0(df::Real, C::Cholesky)
     h_df = df / 2
     p = size(C, 1)
-    h_df * (logdet(C) + p * float(typeof(df))(logtwo)) + logmvgamma(p, h_df)
+    -h_df * (logdet(C) + p * float(typeof(df))(logtwo)) - logmvgamma(p, h_df)
```
mohamed82008, line 51: Why flip the sign here and then outside again?

devmotion:
Yes, and a fix of the variance computation, it was incorrect before. The test error with Zygote is due to the `@warn` in https://github.com/JuliaStats/Distributions.jl/blob/20c91d9efcc5f96913bf8e38be2e3fb14b21942b/src/matrix/wishart.jl#L53. Needs a custom adjoint I guess.

devmotion:
> I don't quite understand how that makes sense from a mathematical point of view.

I guess the name `logc0` is more descriptive, since it's the logarithm of the scalar multiplicative factor. Mathematically it just doesn't matter if you use the variable for `-log(c0)` or `log(c0)`, where `c0` is the previous variable :man_shrugging: I guess one motivation might be that you then avoid the subtractions in the computations of the `logpdf`.

src/matrixvariate.jl
```diff
@@ -124,45 +124,56 @@ end
...
+    R = Base.promote_eltype(T, logc0)
+    prom_S = convert(AbstractArray{T}, S)
+    Wishart{R, typeof(prom_S), typeof(rnk)}(R(df), prom_S, R(logc0), rnk, singular)
+end
+
+_warn(msg) = @warn(msg)
+
+ZygoteRules.@adjoint function _warn(msg)
+    return _warn(msg), _ -> nothing
+end
```
devmotion, line 163: Is there a better solution than just copying the constructor?

johnczito:
> So I haven't really checked the changes closely next to Distributions.jl but this PR seems to pretty much just rename `-c0` to `logc0`. I don't quite understand how that makes sense from a mathematical point of view. Are you sure this is correct?

Sorry! I suppose I'm to blame for these changes. A density is kernel x normalizing constant, which you can write either as (i) f(x) = c₀⋅k(x) or (ii) f(x) = k(x) / Z₀. Doesn't matter which you pick as long as you pick one. The 5 other matrix-variate distributions went with (i) but the `Wishart`s went with (ii). I just harmonized it so that there is a single method `_logpdf(d, X) = logkernel(d, X) + d.logc0` for all of them.

While we're at it (and re [this](https://github.com/TuringLang/DistributionsAD.jl/issues/65#issuecomment-620752798)), I am curious why these distributions have to be line-for-line recoded here. Is it the use of `PDMats`? Where is `Distributions` dropping the ball?

devmotion:
> While we're at it (and re this), I am curious why these distributions have to be line-for-line recoded here. Is it the use of PDMats

I haven't implemented the version in DistributionsAD (just trying to fix it), but yes, it seems, that the major difference is that `Turing(Inverse)Wishart` doesn't work with `AbstractPDMat`s. Otherwise it's mainly minor differences such as additional calls of `float` when converting the constants (not sure why that's needed though).

src/matrixvariate.jl
mohamed82008: I don't think so.

mohamed82008:
For Zygote, we need to avoid stuff that break Zygote like try-catch we re-implement the function avoiding those things that break Zygote. PDMats also causes issues when trying to differentiate wrt the random variable for almost all AD backends. So we have TuringX distributions which avoid PDMats. This is also true for TuringDenseMvNormal.

src/matrixvariate.jl
```diff
@@ -124,45 +124,56 @@ end
...
-
 function unwhiten!(C::Cholesky, x::StridedVecOrMat)
     cf = C.U
     lmul!(transpose(cf), x)
 end
 
+## Custom adjoint since Zygote can't differentiate through `@warn`
+
+ZygoteRules.@adjoint function Wishart(df::T, S::AbstractPDMat{T}, warn::Bool = true) where T
+    return ZygoteRules.pullback(_Wishart, df, S, warn)
```
mohamed82008, line 140: Let's use a do block and define the function inside instead of outside.

devmotion: The reason why I didn't use anonymous functions is that it makes debugging more difficult and is therefore advised against in the ChainRules docs.

mohamed82008: I don't agree that it makes debugging more difficult. In this case, you would still get a useful stacktrace. It's just better to avoid additional function names if it is used in only one place.

devmotion: Hmm I'm not sure, the definition of `_Wishart` is quite long, so I guess it would be easier to keep it separate. Moreover, IMO it is easier to spot that it's just a copy of `Distributions.Wishart` with `@warn` replaced with `_warn`.

devmotion: @johnczito Could we just remove the warning in Distributions, as suggested in https://github.com/JuliaStats/Distributions.jl/pull/1097#discussion_r436377048?

devmotion: I just wanted to update the code and switch to a do block because it doesn't really matter to me in which way it's implemented, but I noticed that type annotations are not possible in do blocks (at least it failed when I tried a simple example with `map`). That means we would have to deviate from the original constructor in Distributions more drastically by removing the type annotations and retrieving the types inside of the function body instead - should we do that?

mohamed82008: I think we can just use the type parameters from the `@adjoint` function header.

devmotion:
CI failure seems to be unrelated, the test was cancelled without any test failures.

mohamed82008:
Thanks @devmotion !

devmotion:
Is it OK to make a new release with this fix (it seems it should be a breaking 0.6 release)? Then we could remove all these temporary fixes in the downstream packages.

mohamed82008:
Sure, sounds good.

devmotion:
Thanks!

Issue #86, opened 2020-06-22
logpdf_with_trans different for Zygote and Forward/ReverseDiff in Dirichlet case


paschermayr:
Hi guys, 

I am using Bijectors v0.8.0, and updated all modules used below yesterday. 
```
  [76274a88] Bijectors v0.8.0
  [163ba53b] DiffResults v1.0.2
  [31c24e10] Distributions v0.23.4
  [ced4e74d] DistributionsAD v0.6.0
  [f6369f11] ForwardDiff v0.10.10
  [37e2e3b7] ReverseDiff v1.2.0
  [e88e6eb3] Zygote v0.4.21

```

I am getting different gradient results depending on whether I use Zygote or Forward/ReverseDiff for the Dirichlet distribution in the `logpdf_with_trans` calculation, MWE below: 

```
using Distributions, DistributionsAD, Bijectors
using ForwardDiff, ReverseDiff, Zygote

function get_loglik(data)
    function loglik(theta_transformed::AbstractVector{<:Real})
    p = Bijectors.invlink(Dirichlet(3,3), theta_transformed)
    lprior = logpdf_with_trans( Dirichlet(3,3), p, true ) ## !!! Difference occurs here
    llik = logpdf(Categorical(p), data)
    return llik + lprior
    end
end
theta_transformed = randn(3)
ll = get_loglik(3)
ll(theta_transformed)

ForwardDiff.gradient(ll, theta_transformed)
ReverseDiff.gradient(ll, theta_transformed) #same as ForwardDiff
Zygote.gradient(ll, theta_transformed) #different than both

```

I am unsure if this should be posted to Bijectors, Distributions, DistributionsAD or Zygote, but started here :D. Apologies if this is not related to Bijectors. 

Best regards, 
Patrick

devmotion:
Found the culprit: the custom adjoints in https://github.com/TuringLang/DistributionsAD.jl/blob/696fe18923a98c3060200455dd059d48d60c3a8d/src/multivariate.jl#L83 and in https://github.com/TuringLang/DistributionsAD.jl/blob/696fe18923a98c3060200455dd059d48d60c3a8d/src/multivariate.jl#L88 are incorrect, there's a `./ x` missing in the third element of the tuple returned by the pullback (that's the computation with respect to the sample).

devmotion:
I'll open a PR with a fix.

torfjelde:
[...]

PR #87, opened 2020-06-22
Fix custom adjoint of `simplex_logpdf`

devmotion:
Fixes https://github.com/TuringLang/DistributionsAD.jl/issues/86.

(BTW I can't stand these testing times, I'll copy the fixes from Bijectors to DistributionsAD.)

### DIFF
--- a/src/multivariate.jl
+++ b/src/multivariate.jl
@@ -80,12 +80,12 @@ function simplex_logpdf(alpha, lmnB, x::AbstractMatrix)
 end
 
 ZygoteRules.@adjoint function simplex_logpdf(alpha, lmnB, x::AbstractVector)
-    simplex_logpdf(alpha, lmnB, x), Δ -> (Δ .* log.(x), -Δ, Δ .* (alpha .- 1))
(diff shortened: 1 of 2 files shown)


devmotion:
I don't know why Zyogte tests failed on macOS - there's no log available, not even in the log archive. Dirichlet tests on linux passed, according to the logs.

torfjelde:
On my screen they're still running?

devmotion:
Hmm maybe I restarted them accidentally? I rechecked the logs that I downloaded, and they are for this PR and from today...

torfjelde:
Seems good to go?

devmotion:
The modified tests passed as well in https://github.com/TuringLang/DistributionsAD.jl/pull/88.

devmotion:
I guess it's fine if I make a new bugfix release? The only change from the previous release would be this fix.

torfjelde:
I'd say so, but I'm not the "main-maintainer" of DistributionsAD, soooo. @mohamed82008 mby?

PR #88, opened 2020-06-22
Simplify tests and do not use `eval`

devmotion:
This PR is basically a slightly modified copy of https://github.com/TuringLang/Bijectors.jl/pull/116. Instead of FiniteDiff it uses FiniteDifferences (since DistributionsAD uses it already), and I added some additional transformations to ensure that parameters and samples are always in the correct domains.

This PR is based on https://github.com/TuringLang/DistributionsAD.jl/pull/87 since it seemed weird to keep the special non-randomized tests for `Dirichlet` that couldn't capture the bug in https://github.com/TuringLang/DistributionsAD.jl/issues/86.


devmotion:
Zygote tests passed (after 30 minutes instead of 1h 48min on master :smiley:), which indicates that https://github.com/TuringLang/DistributionsAD.jl/pull/87 is fine (Zygote tests there are still running).

torfjelde:
So this looks all good to me, but that's because I already saw the same thing earlier :sweat_smile: Should I just accept it or should maybe someone else also have a look, _just_ to be sure?

devmotion:
I requested a review from @mohamed82008 as well since I assume he implemented the current test setup.

yebai:
@mohamed82008

yebai:
thanks, @devmotion -- nice improvements!

mohamed82008:
Sorry guys I was moving houses and had internet issues. Had a quick scan and looks good though.

test/ad/utils.jl
```diff
@@ -0,0 +1,163 @@
...
+                result = logpdf.(dist, xtilde)
+            else
+                result = logpdf(dist, xtilde)
+            end
+            return sum(result)
+        end
+    end
+
+    if isempty(θ)
+        # In this case we can only test the gradient with respect to `x`
```
mohamed82008, line 109: We should test wrt to `x` only even if theta is not empty.

mohamed82008: And we should check if the support is continuous first.

devmotion: The logic is just copied from the previous implementation. There we test wrt `x` if theta is empty and don't check if the support is continuous (see e.g. https://github.com/TuringLang/DistributionsAD.jl/blob/master/test/test_utils.jl#L126 - somehow I forgot to delete the file...). I guess it's OK to not check if the support is continuous, since otherwise we couldn't test any derivatives at all and the `DistSpec` would not be useful. Additionally, both in the old and new implementation we test wrt `x` if theta is not empty and the support is continuous.

mohamed82008: Even if the old implementation didn't test wrt x only with a non-empty theta, we still should. This is useful for prior distributions in Turing.

devmotion: Hmm I can't follow, why would we want to remove tests? As long as tests pass in a reasonable amount of time (which they do with the new implementation), it should always be better to perform more tests, shouldn't it?

mohamed82008: Yes I am suggesting performing more tests :) Currently I don't think we test wrt to `x` and only `x` when theta is not empty.

devmotion: Ah I see, somehow I didn't get it and assumed you would want to remove some of the existing tests :laughing:

Issue #91, opened 2020-06-25
MvNormal broken with Tracker and ReverseDiff


devmotion:
Copied from https://github.com/TuringLang/Bijectors.jl/issues/119#issuecomment-649729101:

I noticed the following
```julia
julia> using DistributionsAD, Distributions, FiniteDifferences, ForwardDiff, Tracker, ReverseDiff, Zygote

julia> function h(theta, x)
           dist = MvNormal(theta)
           return logpdf(dist, x)
       end
h (generic function with 1 method)

julia> x = randn(2)
2-element Array{Float64,1}:
 -0.2728447604886671
  0.2954077831645062

julia> FiniteDifferences.grad(central_fdm(5, 1), x -> h([0.1, 0.1], x), x)[1]
2-element Array{Float64,1}:
  27.28447604886685
 -29.540778316450915

julia> ForwardDiff.gradient(x -> h([0.1, 0.1], x), x)
2-element Array{Float64,1}:
  27.284476048866708
 -29.540778316450616

julia> Tracker.gradient(x -> h([0.1, 0.1], x), theta)[1]
Tracked 2-element Array{Float64,1}:
  2728.4476048866704
 -2954.077831645061

julia> ReverseDiff.gradient(x -> h([0.1, 0.1], x), x)
2-element Array{Float64,1}:
  2728.4476048866704
 -2954.077831645061

julia> Zygote.gradient(x -> h([0.1, 0.1], x), x)[1]
2-element Array{Float64,1}:
  27.284476048866708
 -29.54077831645062
```
with
```julia
  [31c24e10] Distributions v0.23.4
  [26cc04aa] FiniteDifferences v0.10.2
  [f6369f11] ForwardDiff v0.10.10
  [37e2e3b7] ReverseDiff v1.2.0
  [9f7883ad] Tracker v0.2.7
  [e88e6eb3] Zygote v0.5.1
```
Moreover, I get
```julia
julia> FiniteDifferences.grad(central_fdm(5, 1), x -> h([10.0, 10.0], x), x)[1]
2-element Array{Float64,1}:
  0.0027284476047971617
 -0.0029540778316831193

julia> ForwardDiff.gradient(x -> h([10.0, 10.0], x), x)
2-element Array{Float64,1}:
  0.0027284476048866713
 -0.002954077831645062

julia> Tracker.gradient(x -> h([10.0, 10.0], x), theta)[1]
Tracked 2-element Array{Float64,1}:
[...]

PR #92, opened 2020-06-25
Test AD wrt `x` and fixed `theta` if possible

devmotion:
@mohamed82008 noticed that both the previous and the updated test sets (the logic was unchanged) don't test AD with respect to samples `x` and fixed distribution parameters `theta` if `theta` is not empty (only the case when `theta` was empty was covered). This PR adds these additional tests.

Actually, in my local runs some of these tests seem to be broken (see, e.g., https://github.com/TuringLang/DistributionsAD.jl/issues/91) and fail currently. I'm a bit surprised that this hasn't shown up in the other tests with at least one non-fixed distribution parameter.

Somehow the automatic code formatting in VSCode introduced some additional changes...

### DIFF
--- a/test/ad/utils.jl
+++ b/test/ad/utils.jl
@@ -2,7 +2,7 @@
 const AD = get(ENV, "AD", "All")
 
 # Struct of distribution, corresponding parameters, and a sample.
-struct DistSpec{VF <: VariateForm,VS <: ValueSupport,F,T,X,G}
+struct DistSpec{VF<:VariateForm,VS<:ValueSupport,F,T,X,G}
     name::Symbol
     f::F
     "Distribution parameters."
@@ -92,7 +92,7 @@ function test_ad(dist::DistSpec; kwargs...)
     g = dist.xtrans
 
     # Create function with all possible arguments
-    f_allargs = let f = f, g = g
+    f_allargs = let f=f, g=g
         function (x, θ...)
             dist = f(θ...)
             xtilde = g === nothing ? x : g(x)
@@ -112,8 +112,8 @@ function test_ad(dist::DistSpec; kwargs...)
             xtest = mapreduce(vcat, inds) do i
                 vectorize(θ[i - 1])
             end
-            ftest = let xorig = x, θorig = θ, inds = inds
-                x->f_allargs(unpack(x, inds, xorig, θorig...)...)
+            ftest = let xorig=x, θorig=θ, inds=inds
+                x -> f_allargs(unpack(x, inds, xorig, θorig...)...)
             end
             test_ad(ftest, xtest; kwargs...)
         end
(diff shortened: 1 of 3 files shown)


devmotion:
Due to https://github.com/TuringLang/DistributionsAD.jl/issues/91 it comes as no surprise that exactly the Tracker and ReverseDiff tests are failing.

devmotion:
BTW this bug was quite severe and basically affects everyone who used `MvNormal` distributions with fixed parameters and diagonal covariance matrix (both the homoscedastic and heteroscedastic case) and Tracker or ReverseDiff.

devmotion:
It seems tests pass now, it would be great to make a bugfix release as soon as possible :smiley:

devmotion:
Ah wait a second, I think I know why Zygote is still broken in the linked PR (even though it doesn't show up in the tests here...). Let me check that quickly.

devmotion:
The latest commit fixes the linked PR completely, so I guess it's fine if tests pass.

test/ad/utils.jl
```diff
@@ -2,7 +2,7 @@
 const AD = get(ENV, "AD", "All")
 
 # Struct of distribution, corresponding parameters, and a sample.
-struct DistSpec{VF<:VariateForm,VS<:ValueSupport,F,T,X,G}
+struct DistSpec{VF <: VariateForm,VS <: ValueSupport,F,T,X,G}
```
torfjelde, line 5: I think our style-guide says these `<:` should be without the whitespace? Also, there should be whitespace between the type-parameters, e.g. `F,T` should be `F, T`.

test/ad/utils.jl
```diff
@@ -105,58 +105,51 @@ function test_ad(dist::DistSpec; kwargs...)
...
+    for inds in powerset(2:(length(θ) + 1))
+        # Test only distribution parameters
+        if !isempty(inds)
             xtest = mapreduce(vcat, inds) do i
                 vectorize(θ[i - 1])
             end
-            ftest = let xorig=x, θorig=θ, inds=inds
-                x -> f_allargs(unpack(x, inds, xorig, θorig...)...)
+            ftest = let xorig = x, θorig = θ, inds = inds
+                x->f_allargs(unpack(x, inds, xorig, θorig...)...)
```
torfjelde, line 116: I'm not sure if we have a style-guide for anonymous functions, but don't we usually do ` -> ` instead of `->`?

test/ad/utils.jl
```diff
@@ -105,58 +105,51 @@ function test_ad(dist::DistSpec; kwargs...)
...
-                    x -> f_allargs(unpack(x, inds, xorig, θorig...)...)
-                end
-                test_ad(ftest, xtest; kwargs...)
+        # Test derivative with respect to location `x` as well
+        # if the distribution is continuous
+        if Distributions.value_support(typeof(dist)) === Continuous
+            xtest = isempty(inds) ? vectorize(x) : vcat(vectorize(x), xtest)
+            push!(inds, 1)
+            ftest = let xorig = x, θorig = θ, inds = inds
+                x->f_allargs(unpack(x, inds, xorig, θorig...)...)
```
torfjelde, line 127: Same comment regarding style of anonymous functions

devmotion:
Yeah, the style changes were not on purpose, I was just playing around with VSCode and apparently the automatic formatting there (I guess they use JuliaFormatter?) screwed it up.

PR #93, opened 2020-06-26
Broken tests now pass

harryscholes:
These tests now pass for me on all backends


devmotion:
Great, thanks! Yeah, the current test setup is quite unfortunate in that it doesn't check that these tests are actually broken (see also https://github.com/TuringLang/Bijectors.jl/pull/116#discussion_r443399529).

PR #95, opened 2020-06-30
Require backends

nmheim:
This PR tries to separate the different AD backends into files that can be loaded optionally via `Requires.jl`.

## TODO
- [x] univariate distributions
- [x] multivariate distributions
- [x] matrix variate distributions
- [x] product distributions
- [x] test each backend separately
- [ ] ForwardDiff tests for `test/others.jl`
- [x] move adjoints in `zygote.jl` back to original functions definitions


src/DistributionsAD.jl
```diff
@@ -49,14 +43,39 @@ export TuringScalMvNormal,
...
+
+    @require ReverseDiff = "37e2e3b7-166d-5795-8a7a-e32c996b4267" begin
+        include("reversediff.jl")
+    end
+
+    @require Tracker="9f7883ad-71c0-57eb-9f7f-b5c9e6d3789c" begin
+        using DiffRules
+        using ZygoteRules
+        using SpecialFunctions
+        using LinearAlgebra: AbstractTriangular
```
devmotion, line 68: I guess you have to nest dependencies here (see https://github.com/JuliaPackaging/Requires.jl#multiple-requirements), since they are/should not be dependencies of DistributionsAD. Unfortunately, currently some adjoints for Tracker use the pullback of Zygote internally (I assume for performance reasons, I haven't worked on this implementation), indicated by the use of ZygoteRules here. However, that means that users will get cryptic error messages if they don't load Zygote (see e.g. https://github.com/TuringLang/Bijectors.jl/issues/119#issuecomment-649639106) - maybe we could try to remove the dependency on Zygote(Rules) all together and rewrite these adjoints in a different, hopefully not too inefficient way.

src/DistributionsAD.jl
```diff
@@ -49,14 +43,39 @@ export TuringScalMvNormal,
...
+        using Tracker: Tracker, TrackedReal, TrackedVector, TrackedMatrix,
```
devmotion, line 69: Requires requires to load the optional dependencies with a `.` in front of the name (see https://github.com/JuliaPackaging/Requires.jl):
```suggestion
        using .Tracker: Tracker, TrackedReal, TrackedVector, TrackedMatrix,
```
Otherwise users will get a lot of warnings.

src/DistributionsAD.jl
```diff
@@ -49,14 +43,39 @@ export TuringScalMvNormal,
...
 include("matrixvariate.jl")
 include("flatten.jl")
 include("arraydist.jl")
 include("filldist.jl")
-@init @require ReverseDiff = "37e2e3b7-166d-5795-8a7a-e32c996b4267" begin
-    include("reversediff.jl")
+=#
+function __init__()
+    @require ForwardDiff="f6369f11-7733-5829-9624-2563aa707210" begin
+        using ForwardDiff: @define_binary_dual_op # Needed for `eval`ing diffrules here
```
devmotion, line 56:
```suggestion
        using .ForwardDiff: @define_binary_dual_op # Needed for `eval`ing diffrules here
```
See my other comment about `.`

src/DistributionsAD.jl
```diff
@@ -49,14 +43,39 @@ export TuringScalMvNormal,
...
+        using Tracker: Tracker, TrackedReal, TrackedVector, TrackedMatrix,
+                       TrackedArray, TrackedVecOrMat, track, @grad, data
+        include("tracker.jl")
+    end
+
+    @require Zygote="e88e6eb3-aa80-5325-afca-941959d7151f" begin
+        using ForwardDiff
+        using Zygote: @adjoint, pullback
+        include("zygote.jl")
+    end
```
devmotion, line 78: We should not introduce a Requires block for Zygote (in general, Requires is not great and increases loading times a lot, so it's good if we can avoid it). ZygoteRules contains `@adjoint` and `pullback` (they are just reexported by Zygote), and it's a very lightweight package specifically designed so that packages can write custom adjoints but don't have to depend on Zygote or use Requires (same approach as ChainRulesCore, basically). So we should just load `zygote.jl` outside of the `__init__` function in the regular way.

The only problem is that again some custom adjoints call into a different AD backend (here ForwardDiff) that users have to load to be able to use these adjoints. I'm still not sure what's the best way to solve this issue (I was speculating about some dummy methods that display an informative error message but could be redefined/changed when ForwardDiff is loaded).

In a second separate PR one could try to migrate from ZygoteRules to ChainRulesCore (since Zygote supports ChainRules), but that should not be done in this PR.

src/reversediff.jl
```diff
@@ -2,23 +2,29 @@ module ReverseDiffX
 
 export NotTracked
 
-using MacroTools, LinearAlgebra, ..ReverseDiff, StaticArrays
+using LinearAlgebra
+using ..ReverseDiff
+using StaticArrays
+using ZygoteRules
+using Distributions
+using PDMats
```
devmotion, line 10: Are these packages guaranteed to be available? Are they needed for the other packages? Maybe it would be good to move all imports to the Requires blocks for consistency reasons and to spot easier what dependencies are needed and loaded.

src/tracker.jl
```diff
@@ -0,0 +1,329 @@
...
+
+## SpecialFunctions ##
+
+SpecialFunctions.logabsgamma(x::TrackedReal) = track(logabsgamma, x)
+@grad function SpecialFunctions.logabsgamma(x::Real)
+    return logabsgamma(data(x)), Δ -> (digamma(data(x)) * Δ[1],)
+end
+ZygoteRules.@adjoint function SpecialFunctions.logabsgamma(x::Real)
+    return logabsgamma(x), Δ -> (digamma(x) * Δ[1],)
+end
```
devmotion, line 178: This is definitely not needed anymore - there exist adjoints for `logabsgamma` in ChainRules (should be moved to SpecialFunctions at some point, but that's just for now) and Zygote supports ChainRules, so there should be no need for a dedicated adjoint for Zygote.

devmotion:
```suggestion
```

src/tracker.jl
```diff
@@ -0,0 +1,329 @@
...
+    info = data(factors_info[2])
+    return Cholesky{eltype(factors), typeof(factors)}(factors, 'U', info)
+end
+function LinearAlgebra.cholesky(A::Symmetric{<:Any, <:TrackedMatrix}; check=true)
+    uplo = A.uplo == 'U' ? (:U) : (:L)
+    factors_info = symm_turing_chol(parent(A), check, uplo)
+    factors = factors_info[1]
+    info = data(factors_info[2])
+    return Cholesky{eltype(factors), typeof(factors)}(factors, 'U', info)
+end
```
devmotion, line 124: Good to keep in mind that these lines (and many more in this file and in other files for other AD backends) should be moved upstream and are type piracy.

test/runtests.jl
```diff
@@ -12,8 +12,8 @@ using Zygote
 using Random, LinearAlgebra, Test
 
 using Distributions: meanlogdet
-using DistributionsAD: TuringUniform, TuringMvNormal, TuringMvLogNormal,
-    TuringPoissonBinomial
+using DistributionsAD: TuringUniform, #TuringMvNormal, TuringMvLogNormal,
+                       TuringPoissonBinomial
 using ForwardDiff: Dual
```
devmotion, line 17: I can't comment on the lines above, but the import of the AD backends (and hence the whole test setup) has to be changed. Currently all backends (ForwardDiff, ReverseDiff, Tracker, and Zygote) are loaded in this file, meaning that all optional adjoints will be loaded. Hence it is not tested if the optional AD support is actually separated properly.

Instead we have to make sure that only the optional dependencies for the Requires block are loaded that we want to test. One possibility would be to separate the AD tests into separate modules in separate files that are loaded depending on the AD backend, either manually or by using https://github.com/YingboMa/SafeTestsets.jl. More concretely, one would define the general test setup, distributions, etc in a function that takes the tested AD backend (as a symbol?) as an argument and depending on that tests the corresponding backend in `test_ad` instead of relying on the environment variable `AD` there.

src/DistributionsAD.jl
nmheim: Hmm, if we nest the dependencies, the user would have to load all of the by hand, which I though would be a bit cumbersome. If we make them dependencies of DistributionsAD (which is not very nice, I agree) that would not be the case.

We could get rid of the Zygote(Rules) dependency like this:
```julia
 @grad function turing_chol(A::AbstractMatrix, check)                                                                                                                                                                                                                                        
    C = cholesky(data(A), check=check)                                                                                                                                                                                                                                                      
    L = C.L                                                                                                                                                                                                                                                                                 
    ϕ(A::AbstractMatrix) = UpperTriangular(triu(A,1)) .+ Diagonal(A) ./ 2                                                                                                                                                                                                                   
    function back(L̄)                                                                                                                                                                                                                                                                        
        P = ϕ(L̄*L)                                                                                                                                                                                                                                                                          
        L1 = inv(L)                                                                                                                                                                                                                                                                         
        Ā = ϕ(L1' * (P .+ P') * L1)                                                                                                                                                                                                                                                         
        (Ā,nothing)                                                                                                                                                                                                                                                                         
    end                                                                                                                                                                                                                                                                                     
    return (C.factors, C.info), Δ -> back(data(Δ[1]))                                                                                                                                                                                                                                       
end  
```
Code based on [this](https://homepages.inf.ed.ac.uk/imurray2/pub/16choldiff/choldiff.pdf)

src/DistributionsAD.jl
nmheim: Yes, I agree, I introduced that block because of the ForwardDiff dependency. The only adjoint that needs it is the one of [`poissont_pdf_fft`](https://github.com/TuringLang/DistributionsAD.jl/blob/809bdc7ba7ee46363cfa8b226e9d5b2598469056/src/univariate.jl#L287-L294). There is some code that is (or maybe was?) no working with Zygote below the part that I linked.

src/reversediff.jl
nmheim: The ReverseDiff part was already separated from the rest of DAD, but contained its own `ReverseDiffX` package which loads a bunch of dependencies. This part is just reformatted by me (sorry for that), but all these dependencies are guaranteed to be available, yes.

nmheim: Do you think it makes sense to refactor the ReverseDiff part or should we keep it like this for now?

src/DistributionsAD.jl
devmotion: It seems ChainRules (not ChainRulesCore) actually contains some implementations of this paper for the reverse-mode case: https://github.com/JuliaDiff/ChainRules.jl/blob/224e553132f26979e7f5c34fccaf769788022299/src/rulesets/LinearAlgebra/factorization.jl#L72-L83

This doesn't solve the issues here with mixed AD but it seems that all the special adjoints and workarounds for Cholesky with ZygoteRules/Zygote might not be needed at all for the Zygote backend. One idea might be to depend on ChainRules for now (although ideally we would only want to depend on ChainRulesCore), and then call `rrule` here explicitly, without having to rely on Zygote. I still think it's weird that users should have to load both AD backends for these adjoints.

test/runtests.jl
nmheim: Yes, I very much agree! :) I'll take a look at SafeTestsets

src/reversediff.jl
devmotion: Ah I think the reason for this additional module was that all/most of this code should actually go into ReverseDiff and moved upstream. Maybe leave it in a separate module for now (so it's more clearly separated), but put everything in one file? In any case, this is not a critical issue I guess.

src/DistributionsAD.jl
nmheim: Yes, I also think they should only have to load `Tracker` here. I could copy the reverse functions from ChainRules and use them in the `@grad` adjoint for Tracker?

devmotion: I usually try to avoid code duplications, if not needed. IMO it would be fine to add a general dependency on ChainRules (its dependencies don't seem to be too bad, see https://github.com/JuliaDiff/ChainRules.jl/blob/224e553132f26979e7f5c34fccaf769788022299/Project.toml#L5-L10) instead of copying the code. Then these rules and adjoints are guaranteed to be available for every backend, and hopefully we could call `rrule` here directly without having to rely on Zygote (see https://www.juliadiff.org/ChainRulesCore.jl/stable/#Example-of-using-ChainRules-directly.-1 for some examples).

nmheim: I thought `Tracker` does not support `ChainRules` yet?

nmheim: Ahh, yes, sorry, we can just call the `rrule` directly. ofc :)

devmotion: Yes, but that's a different use case here. Tracker doesn't support it just means does Tracker can't process the custom adjoints defined with ChainRulesCore. However, you could still just call `rrule(cholesky, ...)` inside of the custom adjoint for Tracker and would get the function and its pullback, similar to what is currently done with ZygoteRules/Zygote.

devmotion: Ah I saw your last reply only after I had sent my comment :)

src/DistributionsAD.jl
nmheim: I have removed the Zygote require block, as suggested and moved the `@adjoint` for the `poissonbinomial_pdf_fft` to `zygote_forwarddiff.jl` which is in the ForwardDiff require block. Note that all the ForwardDiff code will be loaded together with Zygote, because Zygote depends on ForwardDiff... so the dummy `@adjoint` in `zygote.jl` will practically never be hit.

src/DistributionsAD.jl
nmheim: My attempt for both `Tracker` and `ReverseDiff` is in `common.jl`, `tracker.jl`, and `reversediffx.jl`

test/runtests.jl
nmheim: I tried separating the backend tests in a similar way as it is done in `tests/ad/utils.jl`. Not sure if its the best way though...

src/DistributionsAD.jl
nmheim: Actually, I think we could just move the zygote adjoints back to the `univaritate.jl` file as it is not loaded conditionally any more

src/common.jl
```diff
@@ -1,225 +1,53 @@
...
-    if C.uplo == 'U'
-        return copy(upper(C.factors)')
-    else
-        return lower(C.factors)
+function turing_chol_back(A::AbstractMatrix, check)
+    C, dC_pullback = rrule(turing_chol, A, check)
+    function back(Δ)
+        _, dC = dC_pullback(Δ)
+        dC = unthunk(dC)
+        (dC, nothing)
```
devmotion, line 22:
```suggestion
        return (unthunk(dC), nothing)
```

src/common.jl
```diff
@@ -1,225 +1,53 @@
...
-symm_turing_chol(A::TrackedMatrix, check, uplo) = track(symm_turing_chol, A, check, uplo)
-@grad function symm_turing_chol(A::AbstractMatrix, check, uplo)
-    C, back = ZygoteRules.pullback(data(A), check, uplo) do A, check, uplo
-        cholesky(Symmetric(A, uplo), check=check)
+function symm_turing_chol_back(A::AbstractMatrix, check, uplo)
+    C, dC_pullback = rrule(symm_turing_chol, A, check, uplo)
+    function back(Δ)
+        _, dC = dC_pullback(Δ)
+        dC = unthunk(dC)
+        (dC, nothing, nothing)
```
devmotion, line 46:
```suggestion
        return (unthunk(dC), nothing, nothing)
```

src/DistributionsAD.jl
devmotion: I think it's better to keep them in separate files, regardless if they're loaded conditionally or not. 

Regarding `poissonbinomial_pdf_fft`, I guess you could indeed just add a comment to the non-dummy version that explains that Zygote loads ForwardDiff, and hence it will always be available for Zygote. Maybe also add a comment in the Zygote file, that actually one adjoint is defined in the optional ForwardDiff code.

test/ad/distributions.jl
```diff
@@ -18,14 +18,18 @@
     to_posdef_diagonal(a::AbstractVector) = Diagonal(a.^2 .+ 1)
 
     # Define adjoints for Tracker
-    to_posdef(A::TrackedMatrix) = Tracker.track(to_posdef, A)
-    Tracker.@grad function to_posdef(A::TrackedMatrix)
-        data_A = Tracker.data(A)
-        S = data_A * data_A' + I
-        function pullback(∇)
-            return ((∇ + ∇') * data_A,)
+    if AD == "All" || AD == "ForwardDiff_Tracker"
```
devmotion, line 21: I guess we should split `ForwardDiff_Tracker` in `ForwardDiff` and `Tracker`

test/ad/distributions.jl
```diff
@@ -18,14 +18,18 @@
...
+        @eval begin
+            to_posdef(A::TrackedMatrix) = Tracker.track(to_posdef, A)
+            Tracker.@grad function to_posdef(A::TrackedMatrix)
+                data_A = Tracker.data(A)
+                S = data_A * data_A' + I
+                function pullback(∇)
+                    return ((∇ + ∇') * data_A,)
+                end
+                return S, pullback
+            end
```
devmotion, line 31: Why is `@eval` needed here but not below?

test/others.jl
```diff
@@ -109,89 +116,102 @@
...
+
+                # Check that ReverseDiff forwards-pass produces the correct answer.
+                @test isapprox(y, y_rd, atol=atol, rtol=rtol)
+
+                # Check that ReverseDiff reverse-mode sensitivities are correct.
+                @test isapprox(x̄s_rd, x̄s_fdm[1]; atol=atol, rtol=rtol)
+            end
         end
+
+        if AD == "All" || AD == "ForwardDiff_Tracker"
```
devmotion, line 175: Again split ForwardDiff an Tracker tests if possible.

test/runtests.jl
```diff
@@ -3,18 +3,31 @@ using DistributionsAD
...
+elseif AD == "Zygote"
+    @eval using Zygote
+elseif AD == "ReverseDiff"
+    @eval using ReverseDiff
+elseif AD == "ForwardDiff_Tracker"
+    @eval using ForwardDiff
+    @eval using Tracker
+else
+    error("Unknown AD backend: $AD")
+end
```
devmotion, line 24: Maybe just
```suggestion
if AD == "All" || AD == "ForwardDiff"
    @eval using ForwardDiff
end
if AD == "All" || AD == "Zygote"
    @eval using Zygote
end
if AD == "All" || AD == "ReverseDiff"
    @eval using ReverseDiff
end
if AD == "All" || AD == "Tracker"
    @eval using Tracker
end
```

devmotion: BTW we have to add an additional Github action if we split ForwardDiff and Tracker.

test/ad/distributions.jl
nmheim: Yes, I agree, but I wanted to ask first if there was a reason for Tracker/ForwardDiff being tested together in the existing code

devmotion: I guess the reason was mainly that FiniteDifferences wasn't used at all and everything was compared against ForwardDiff (at least that was the status a week ago when I started refactoring the tests IIRC).

test/ad/distributions.jl
nmheim: This is a mistake and actually caused a nasty bug that I was hunting for a while now. Thanks for pointing this out. Is it even possible to overload a function within an if condition somehow?

devmotion: I assume moving the function definitions to a separate file and including it conditionally should work?

test/PosDef.jl
```diff
@@ -0,0 +1,28 @@
+module PosDef
```
devmotion, line 1: Is it necessary to put it in a separate module? I would be surprised at least if the use of Requires and separate modules are needed to define the adjoint for Tracker.

I would have assumed that something like
```julia
if AD == "All" || AD == "Tracker"
    include("tracker_utils.jl")
end
```
in test/ad/distributions.jl would be sufficient.

nmheim: The problem is that the `TrackedMatrix` overload is compiled even if `Tracker` is not loaded...
I tried it with Requires in the `test/ad/distributions.jl` file which didn't work, but defining `is_posdef` in `test/runtests.jl` and requiring the Tracker stuff there does the job. So there is no more separate module, but there might be an obvious solution I am not seeing?

nmheim:
There is a file `src/for.jl` that is not loaded in `src/DistributionsAD.jl` should I remove it?

yebai:
> There is a file src/for.jl that is not loaded in src/DistributionsAD.jl, that was introduced in this PR: #13
> what to do with it?

@nmheim These codes should probably live in the package `Probability.jl`. But for now, perhaps consider moving these code into a dedicated (temporary) subfolder, say `combinators/`?

Project.toml
```diff
@@ -3,14 +3,13 @@ uuid = "ced4e74d-a319-5a8a-b0ac-84af2272839c"
 version = "0.6.2"
 
 [deps]
-Combinatorics = "861a8166-3701-5b0c-9a16-15d98fcdc6aa"
+ChainRules = "082447d4-558c-5d27-93f4-14fc19e9eca2"
```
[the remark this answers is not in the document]
nmheim: First, thanks a lot for all the suggestions! :)

ChainRules is needed for now, because we are using `chol_blocked_rev` here: https://github.com/TuringLang/DistributionsAD.jl/blob/5ec520116e3ff8b6f80534a3029674020c464843/src/common.jl#L11

src/common.jl
```diff
@@ -1,225 +1,49 @@
...
-# Work around https://github.com/FluxML/Tracker.jl/pull/9#issuecomment-480051767
-
-upper(A::AbstractMatrix) = UpperTriangular(A)
-lower(A::AbstractMatrix) = LowerTriangular(A)
-function upper(C::Cholesky)
-    if C.uplo == 'U'
-        return upper(C.factors)
-    else
-        return copy(lower(C.factors)')
+function ChainRules.rrule(::typeof(turing_chol), A::AbstractMatrix, check)
```
sethaxen, line 7: Currently `chol_blocked_rev` is only tested for real matrices. Probably the right thing to do is relax that constraint on ChainRules and test (I'll open a PR).

Project.toml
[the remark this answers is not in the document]
sethaxen: You're welcome! That makes sense. I added an alternate suggestion below.

nmheim:
I will push the suggestions by @sethaxen once ChainRules can deal with keyword arguments. Then this should be ready.

nmheim:
Ok, it will take some more time until the keyword stuff for the `cholesky` rule is fixed. Seth will ping me once thats done. I think we could merge this? @devmotion @mohamed82008

PR #96, opened 2020-07-18
Handle Real and Any better

mohamed82008:
This PR allows the handling of arrays of Any and Real better by relaxing some type constraints.

### DIFF
--- a/src/mvcategorical.jl
+++ b/src/mvcategorical.jl
@@ -77,7 +77,7 @@ function Base.rand(rng::AbstractRNG, d::MvDiscreteNonParametric{T,P}) where {T,P
     p = probs(d)
     n, k = size(p)
     map(1:k) do j
-        draw = rand(rng, P)
+        draw = rand(rng, (P === Real ? Float64 : P))
         cp = zero(P)
         i = 0
         while cp < draw && i < n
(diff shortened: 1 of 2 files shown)


mohamed82008:
Tests are failing because of an issue with the latest ReverseDiff version which has a lot of the functionality in DistributionsAD. This created some conflict. Should be alright after #97 goes in though.

src/mvcategorical.jl
```diff
@@ -77,7 +77,7 @@ function Base.rand(rng::AbstractRNG, d::MvDiscreteNonParametric{T,P}) where {T,P
     p = probs(d)
     n, k = size(p)
     map(1:k) do j
-        draw = rand(rng, P)
+        draw = rand(rng, (P === Real ? Float64 : P))
```
devmotion, line 80: That seems like a quite crude heuristic. Ideally we would want `P` to be
```julia
mapreduce(promote_type, 1:n) do i
    typeof(p[i, j])
end
```
assuming that all provided probabilities can be promoted to a common non-abstract type in each column (but possibly not the same type for all columns) - I guess that should usually be satisfied?

BTW copying the logic from `Distributions.DiscreteNonParametric` here unfortunately introduces also the bugs present in the implementation in Distributions (see e.g. https://github.com/JuliaStats/Distributions.jl/pull/1128), which might not be fixed when the upstream bugs are fixed if people don't keep track of Distributions closely.

PR #109, opened 2020-08-16
Require Julia 1.3

devmotion:
Tracker issues related to PDMats 0.10 are fixed in the recent release of Tracker which, however, is not available on Julia < 1.3. To avoid having to backport fixes to earlier Julia versions, we drop support of Julia < 1.3 and add tests for Julia 1.3 (probably, waiting for more comments on https://discourse.julialang.org/t/rfc-dropping-support-for-julia-1-3-in-turing-jl/45015).

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "DistributionsAD"
 uuid = "ced4e74d-a319-5a8a-b0ac-84af2272839c"
-version = "0.6.3"
+version = "0.6.4"
 
 [deps]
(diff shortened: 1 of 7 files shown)


Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "DistributionsAD"
 uuid = "ced4e74d-a319-5a8a-b0ac-84af2272839c"
-version = "0.6.4"
+version = "0.6.3"
```
devmotion: I guess there's no need to increase the version number - 0.6.3 will work fine on newer Julia versions as well.

PR #111, opened 2020-08-25
Fix TrackedReal nextfloat and prevfloat

avik-pal:
As discussed with @DhairyaLGandhi it would make sense to move these functions to Tracker.jl

### DIFF
--- a/src/tracker.jl
+++ b/src/tracker.jl
@@ -5,11 +5,11 @@ Tracker.dual(x::Int, p) = x
 
 Base.prevfloat(r::TrackedReal) = track(prevfloat, r)
 @grad function prevfloat(r::Real)
-    prevfloat(data(r)), Δ -> Δ
+    prevfloat(data(r)), Δ -> (Δ,)
 end
 Base.nextfloat(r::TrackedReal) = track(nextfloat, r)
 @grad function nextfloat(r::Real)
-    nextfloat(data(r)), Δ -> Δ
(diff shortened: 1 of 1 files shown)


yebai:
It's a good idea to consider transferring `Tracker.jl` specific rules back to `Tracker.jl`, perhaps under a dedicated folder for `Distributions.jl`. More generally, we should consider transferring implementations related to `Zygote.jl`, `ReverseDiff.jl` and `ForwardDiff.jl` back to their main repo as well. 

However, we don't currently have merge access for repos mentioned above. It would be helpful if some Turing members (e.g. @mohamed82008, @willtebbutt, @devmotion and @yebai) are allowed to maintain these repos such that bug fixes can be merged quickly.

willtebbutt:
I would not be pro- this happening. Now that https://github.com/JuliaDiff/ChainRulesCore.jl/pull/182 is in, `Tracker` and `ReverseDiff` should join `Zygote` and move over to `ChainRules`. The `@adjoint`s currently implemented should be replaced with `rrule`s.

mohamed82008:
Ya I agree with Will, ChainRules is the future. May not be worth moving things around now unless they are urgently needed by users who don't want to load DistributionsAD.

mohamed82008 (approved):

yebai:
Maintaining so much AD libraries is probably not the greatest idea in my view. In particular, Tracker and RecerseDiff are basically the same apart from implementation details.

PR #112, opened 2020-08-27
Follow API and documentation of Distributions

devmotion:
Currently DistributionsAD does not adhere to the documentation and API of Distributions. For multivariate and matrixvariate distributions one should implement `Distributions._logpdf` instead of `Distributions.logpdf`. The current implementation leads to issues (reported on Slack) with https://github.com/TuringLang/DynamicPPL.jl/pull/153, which requires a proper implementation of the interface.

I spent surprisingly much time on getting the tests to (somewhat) pass locally, so maybe some additional fixes are needed.


devmotion:
BTW the branch https://github.com/TuringLang/DynamicPPL.jl/tree/fix_loglikelihood contains a commit that reverts the change in DynamicPPL. IMO the correct fix is to change the implementation in DistributionsAD, but it might still be helpful as a current workaround.

devmotion:
The PR does not remove any previously existing functionality (at least all tests still pass) but implements `_logpdf` instead. `logpdf(::Distribution, ::AbstractMatrix)` is only needed since the upstream implementation falls back to an in-place algorithm, IMO that should be changed in Distributions - otherwise it would only be useful if there are more efficient implementations. Similarly, implementations of `loglikelihood` are only useful if there are more efficient ways for evaluating it than looping over all samples.

yebai (approved):

yebai:
Thanks, @devmotion. It looks all good to me!

PR #117, opened 2020-09-24
Adding an implementation of a mean-field MvNormal

theogf:
This PR aims at adding the type `TuringMFMvNormal` for a structured mean-field implementation of the multivariate Gaussian, i.e. for $(x_1, x_2, ...., x_n) = x ~ q(x) = \prod_i^M q_{S_i}(x_{S_i})$ where $S_i$ are $M$ disjoint subsets of $(1,2, ... n)$.
Basically you want some group of variables to be correlated and others to be independent. This is especially useful for variational inference as the inference cost is decreased.
I used the `BlockDiagonal` implementation from [`BlockDiagonals.jl`](https://github.com/invenia/BlockDiagonals.jl) to inforce the independence between group of variables.
There seems to be still some issues with the AD, I will be looking into it

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -4,6 +4,7 @@ version = "0.6.9"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
+BlockDiagonals = "0a1fb500-61f7-11e9-3c65-f5ef3456f9f0"
 ChainRules = "082447d4-558c-5d27-93f4-14fc19e9eca2"
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
 Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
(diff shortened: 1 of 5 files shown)


devmotion:
I'm not sure if a new type should be added to DistributionsAD here - isn't this just a special form of `TuringMvNormal` with a `BlockDiagonal` as covariance matrix? In that case I think it would be better to either let users supply the `BlockDiagonal` as covariance matrix themselves or add constructors for `TuringMvNormal`.

In general, I'm wondering where we want to move with DistributionsAD. We had some discussions a while ago but it is not completely clear what the future goal or outline of this package is. In particular, I think ideally `TuringMvNormal` etc. and all the type piracy shouldn't be needed if Distributions doesn't enforce the covariance matrix of `MvNormal` to be a `AbstractPDMat` - then Distributions (or another package such as DistributionsAD) could just define adjoints etc if needed but wouldn't have to come up with all these additional types.

theogf:
@devmotion Got it! I was starting to implement for AdvancedVI.jl and then I thought it had a better place in DistributionsAD.jl

devmotion:
> Also it allows me to create a update function (as in AVI.jl) :

You could use this implementation even if `TuringMFMvNormal` doesn't exist by dispatching on the type of the covariance matrix of `TuringMvNormal`, couldn't you?

theogf:
No you can't because `TuringDenseMvNormal` is contrained to have a `Cholesky` type for the covariance

PR #124, opened 2020-10-30
Relaxed parametrization of TuringDenseMvNormal and add new constructor

theogf:
The idea is the following:
Right now `TuringDenseMvNormal` takes a covariance matrix `A` and stores its cholesky decomposition `C`, which is in return used to sample : `x = mu + C * z`.
However it would be beneficial to have a constructor that takes directly `X` as in `A = X * X'`, which is often done for variational inference for instance.
Indeed with `X` given we can sample similarly with `x = mu + X * z`, except that 1) we don't need to compute the cholesky, 2) It allows for low rank matrices!

Right now I am a bit confused by the way tests are run, so I don't know how to add this one in the tests

### DIFF
--- a/src/multivariate.jl
+++ b/src/multivariate.jl
@@ -90,17 +90,22 @@ end
 
 A multivariate Normal distribution whose covariance is dense. Compatible with Tracker.
 """
-struct TuringDenseMvNormal{Tm<:AbstractVector, TC<:Cholesky} <: ContinuousMultivariateDistribution
+struct TuringDenseMvNormal{Tm<:AbstractVector, TC<:AbstractMatrix} <: ContinuousMultivariateDistribution
     m::Tm
     C::TC
 end
 function TuringDenseMvNormal(m::AbstractVector, A::AbstractMatrix)
-    return TuringDenseMvNormal(m, cholesky(A))
+    return TuringDenseMvNormal(m, cholesky(A).L)
 end
+
+function TuringSqrtDenseMvNormal(m::AbstractVector, C::AbstractMatrix)
+    TuringDenseMvNormal(m, C)
+end
+
 Base.length(d::TuringDenseMvNormal) = length(d.m)
 Distributions.rand(d::TuringDenseMvNormal, n::Int...) = rand(Random.GLOBAL_RNG, d, n...)
 function Distributions.rand(rng::Random.AbstractRNG, d::TuringDenseMvNormal, n::Int...)
-    return d.m .+ d.C.U' * adapt_randn(rng, d.m, length(d), n...)
+    return d.m .+ d.C * adapt_randn(rng, d.m, length(d), n...)
 end
 
 """


src/multivariate.jl
```diff
@@ -90,17 +90,22 @@ end
...
+
+function TuringSqrtDenseMvNormal(m::AbstractVector, C::AbstractMatrix)
+    TuringDenseMvNormal(m, C)
+end
+
 Base.length(d::TuringDenseMvNormal) = length(d.m)
 Distributions.rand(d::TuringDenseMvNormal, n::Int...) = rand(Random.GLOBAL_RNG, d, n...)
 function Distributions.rand(rng::Random.AbstractRNG, d::TuringDenseMvNormal, n::Int...)
-    return d.m .+ d.C.U' * adapt_randn(rng, d.m, length(d), n...)
+    return d.m .+ d.C * adapt_randn(rng, d.m, length(d), n...)
```
devmotion, line 108: Here `.U'` is used on purpose to avoid a copy. By default `cholesky` computes the upper triangular matrix which can be accessed without any additional allocations - however, `.L` (as in the modified) creates a new array:
https://github.com/JuliaLang/julia/blob/719c4d01b98c9c6a0d531047a7d530eae90e2af9/stdlib/LinearAlgebra/src/cholesky.jl#L445-L457

src/multivariate.jl
```diff
@@ -90,17 +90,22 @@ end
...
     m::Tm
     C::TC
 end
 function TuringDenseMvNormal(m::AbstractVector, A::AbstractMatrix)
-    return TuringDenseMvNormal(m, cholesky(A))
+    return TuringDenseMvNormal(m, cholesky(A).L)
 end
+
+function TuringSqrtDenseMvNormal(m::AbstractVector, C::AbstractMatrix)
+    TuringDenseMvNormal(m, C)
```
devmotion, line 102: I don't see how this could work? It will use the constructor above and perform a Cholesky decomposition anyways?

devmotion:
Is the intention to use (in some sense) arbitrary matrices `X`? Or just lower or upper triangular ones? In the latter case one could just wrap them in a `Cholesky` struct.

src/multivariate.jl
theogf: Or one can just call `cholesky(A, :L)` in the first place https://github.com/JuliaLang/julia/blob/719c4d01b98c9c6a0d531047a7d530eae90e2af9/stdlib/LinearAlgebra/src/cholesky.jl#L406

src/multivariate.jl
theogf: Oh you're right, the constructor is wrong, the following would work directly no ? 
```suggestion
    TuringDenseMvNormal{typeof(m), typeof(C)}(m, C)
```

theogf:
> Is the intention to use (in some sense) arbitrary matrices `X`? Or just lower or upper triangular ones? In the latter case one could just wrap them in a `Cholesky` struct.

Yes exactly, actually `X` does not even need to be a square matrix.

src/multivariate.jl
devmotion: Yes, that could be changed. It might have to be changed in other places as well though and would require to always use `.L` then. I guess it will be surprising if the defaults are different from Julia base - suddenly users will see additional allocations when they use the default factorization (e.g., by computing it yourself with PositiveFactorizations).

devmotion:
I'm curious, when is it useful to parameterize covariance matrices in more general ways than with regular Cholesky factors (or submanifolds of Cholesky factors)? I would have assumed that more flexible parameterizations would introduce identifiability issues. Also it seems that non-square matrices would require different implementations of `rand` (and possibly other methods) since they would lead to dimension mismatches currently?

theogf:
The dimension mismatch is certainly a good comment. Would you be open, that I leave TuringDenseMvNormal alone and create TuringLowRankMvNormal which would take this `X` as an argument? Of course giving a square matrix X would lead to a full rank, like in Dense, just the implementation would be different.

devmotion:
I don't think adding more types is necessarily a good idea.

Maybe it would make sense to allow any type for `mean` and `cov` in a `TuringMvNormal` struct (as `MvNormal` maybe should...) and get rid of `TuringDenseMvNormal`, `TuringScalMvNormal` etc.. The current implementation could instead be based on dispatches of `mean` and `cov` (maybe even with an orthogonal design to more easily allow for a large set of possible combinations of types). Then it would also be easier to add support for custom types in other packages by only defining `rand` and `_logpdf`/`loglikelihood` (or whatever orthogonal interface one would use).

For your use case it would mean that you could just define your own type of low rank factorization and define the corresponding methods, without having to maintain support for all matrix (and vector) types that could possibly be interesting for users and downstream packages.

theogf:
Since it will probably never be merged...

PR #132, opened 2020-11-20
Update tests

devmotion:
This PR copies some fixes of the tests from https://github.com/TuringLang/Bijectors.jl/pull/146. Currently the transformations of the samples are not applied to `filldist` and `arraydist` which can lead to problems for distributions with constrained support.


torfjelde:
> Currently the transformations of the samples are not applied to filldist and arraydist which can lead to problems for distributions with constrained support.

Uhmm, am confused. What do you mean here? That in current #master, we're not testing transformed `filldist` and `arraydist`?

devmotion:
Ah yeah sorry for this bad explanation. This affects only the tests: When we test matrixvariate distributions, then we also test the `filldist` and `arraydist` versions of univariate and multivariate distributions. For some of them (like Dirichlet) samples are constrained to a specific subspace of R^n. To ensure that also slightly perturbed versions of the samples are always in this space, e.g. when working with finite differencing, we do not define only a single valid sample but instead specify a transformation to this space for samples in R^n and use this transformation in our AD tests. However, currently these transformations are not used in the matrixvariate tests with `filldist` and `arraydist`.

torfjelde (approved):
LGTM!

PR #142, opened 2020-12-11
Don't depend on ChainRules internals

### LINKED ISSUE #131: Cholesky internals in ChainRules are changing
https://github.com/JuliaDiff/ChainRules.jl/pull/311 will remove `chol_blocked_rev` which is used here: https://github.com/TuringLang/DistributionsAD.jl/blob/eb36932964ddb24ee591865bce9be94238e1428f/src/common.jl#L11

nmheim:
ChainRules removed the function `ChainRules.chol_blocked_rev` which we used for the `turing_chol` functions. This should fix #131


devmotion (approved):
LGTM if tests pass. Please bump the version as well.

In principle, it would be nice if we would not need any special casing in DistributionsAD. For which AD backends are these workarounds needed? Zygote should already use the definitions in ChainRules but I assume Tracker or ForwardDiff might be problematic?

nmheim:
Yes, they are only used in Tracker and ReverseDiff. I am not that familiar with those packages to get rid of them I am afraid...

devmotion:
Thanks!

PR #151, opened 2021-01-29
Fix Dirichlet with ReverseDiff

devmotion:
This PR fixes the problems with ReverseDiff, tests pass locally both with Distributions 0.24.10 and 0.24.12.

The recent releases of Distributions revealed different bugs in the implementation for ReverseDiff:
- `simplex_logpdf` was not imported in `src/reversediff.jl` and therefore in the ReverseDiff module a new function `simplex_logpdf` was defined which was never reached
- the definition of the pullback for `simplex_logpdf` in `src/reversediff.jl` was incorrect
- dispatches for `logpdf`, `_logpdf`, and `loglikelihood` were missing
- for ReverseDiff an adjoint of `softmax` was missing (and without it the gradients were incorrect - I wonder if recent changes of `softmax` in NNlib broke it)

Some additional simplifications are possible with more recent versions of Distributions (see the other PR) but this PR only contains the fixes for ReverseDiff.


devmotion:
I managed to condense the `view` + `track` problems to a MWE and proposed a fix in ReverseDiff: https://github.com/JuliaDiff/ReverseDiff.jl/pull/164

src/reversediff.jl
```diff
@@ -47,12 +48,25 @@ using ..DistributionsAD: TuringPoissonBinomial,
...
+                         TuringDenseMvNormal,
+                         VectorOfMultivariate,
+                         FillVectorOfMultivariate
 
 include("reversediffx.jl")
 
 adapt_randn(rng::Random.AbstractRNG, x::TrackedArray, dims...) = adapt_randn(rng, value(x), dims...)
 
+# without this definition tests of `VectorOfMultivariate` with `Dirichlet` fail
+# maybe an upstream bug caused by `view`?
```
devmotion, line 60:
```suggestion
# upstream bug caused by `view` + `track`: https://github.com/JuliaDiff/ReverseDiff.jl/pull/164
```

src/reversediff.jl
```diff
@@ -240,36 +254,60 @@ end
...
+end
+function logpdf(d::Dirichlet, x::AbstractMatrix{<:TrackedReal})
+    return logpdf(TuringDirichlet(d.alpha, d.alpha0, d.lmnB), x)
+end
+function loglikelihood(d::Dirichlet, x::AbstractMatrix{<:TrackedReal})
+    return loglikelihood(TuringDirichlet(d.alpha, d.alpha0, d.lmnB), x)
+end
+
+# default definition of `loglikelihood` yields gradients of zero?!
+# again `view` seems to be problematic
```
devmotion, line 273:
```suggestion
# upstream bug caused by `view` + `track`: https://github.com/JuliaDiff/ReverseDiff.jl/pull/164
```

src/reversediff.jl
```diff
@@ -240,36 +254,60 @@ end
...
     @eval $func_header = track(simplex_logpdf, alpha, lmnB, x)
 end
 @grad function simplex_logpdf(alpha, lmnB, x::AbstractVector)
-    simplex_logpdf(value(alpha), value(lmnB), value(x)), Δ -> begin
-        (Δ .* log.(value(x)), -Δ, Δ .* (value(alpha) .- 1))
+    _alpha = value(alpha)
+    _lmnB = value(lmnB)
+    _x = value(x)
+    simplex_logpdf(_alpha, _lmnB, _x), Δ -> begin
+        (Δ .* log.(_x), -Δ, Δ .* (_alpha .- 1) ./ _x)
```
yebai, line 302: Thanks, @devmotion, nice catch. I wonder why our tests miss bugs in the current version?

src/reversediff.jl
```diff
@@ -240,36 +254,60 @@ end
...
     end
 end
 @grad function simplex_logpdf(alpha, lmnB, x::AbstractMatrix)
-    simplex_logpdf(value(alpha), value(lmnB), value(x)), Δ -> begin
-        (log.(value(x)) * Δ, -sum(Δ), repeat(value(alpha) .- 1, 1, size(x, 2)) * Diagonal(Δ))
+    _alpha = value(alpha)
+    _lmnB = value(lmnB)
+    _x = value(x)
+    simplex_logpdf(_alpha, _lmnB, _x), Δ -> begin
+        (log.(_x) * Δ, -sum(Δ), ((_alpha .- 1) ./ _x) * Diagonal(Δ))
```
yebai, line 310: looks correct to me!

yebai (approved):
Thanks, @devmotion - it looks good to me! I left a few clarification questions below.

src/reversediff.jl
```diff
@@ -240,36 +254,60 @@ end
...
+
+# default definition of `loglikelihood` yields gradients of zero?!
+# upstream bug caused by `view` + `track`: https://github.com/JuliaDiff/ReverseDiff.jl/pull/164
+function loglikelihood(d::TuringDirichlet, x::AbstractMatrix{<:TrackedReal})
+    return sum(i -> logpdf(d, x[:, i]), axes(x, 2))
+end
+
 for func_header in [
-    :(simplex_logpdf(alpha::TrackedVector, lmnB::Real, x::AbstractVector)),
+    :(simplex_logpdf(alpha::AbstractVector{<:TrackedReal}, lmnB::Real, x::AbstractVector)),
```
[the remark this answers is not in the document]
devmotion: No, it's just 1) we were not able to test these definitions before since `simplex_logpdf` was not imported in the ReverseDiff module, i.e., they define a local `simplex_logpdf` instead of `DistributionsAD.simplex_logpdf` and 2) that currently `view`s of TrackedArray create an `AbstractArray{<:TrackedReal}` but not a `TrackedArray` and hence this covers more cases (see also the ReverseDiff PR).

src/reversediff.jl
devmotion: The main issue is that the ReverseDiff functionality lives in its own module (I assume to avoid name clashes with Tracker?) and `simplex_logpdf` was not imported, i.e., these definitions were local functions of the ReverseDiff module but not actually DistributionsAD.simplex_logpdf. Thus it was not possible to test them and since both Dirichlet and TuringDirichlet just worked with ReverseDiff the tests did not reveal it.

PR #161, opened 2021-04-19
Fix errors with Adapt 3.3.0

devmotion:
This PR fixes errors with Adapt 3.3.0 (see https://github.com/SciML/OrdinaryDiffEq.jl/issues/1369#issuecomment-817286051 for an explanation why it broke `adapt_randn`).

Additionally, I used the opportunity to replace the custom `rrule` with `@non_differentiable` which now supports varargs (moreover, it defines a `frule` as well and uses an optimized definition with `@nospecialize`).


torfjelde (approved):
LGTM!

Issue #163, opened 2021-04-26
Package Warning


gideonsimpson:
Just started seeing this:

```
┌ Warning: Package DistributionsAD does not have ForwardDiff in its dependencies:
│ - If you have DistributionsAD checked out for development and have
│   added ForwardDiff as a dependency but haven't updated your primary
│   environment's manifest file, try `Pkg.resolve()`.
│ - Otherwise you may need to report an issue with DistributionsAD
│ Loading ForwardDiff into DistributionsAD from project dependency, future warnings for DistributionsAD are suppressed.
└ @ nothing nothing:888
```

devmotion:
Ah, it seems this was introduced accidentally in https://github.com/TuringLang/DistributionsAD.jl/pull/162 where it was possible to remove the dependency on ForwardDiff. It seems the optionally loaded ForwardDiff functionality was not implemented correctly but it was not a problem before since ForwardDiff was actually a proper dependency. And unfortunately it is not possible to notice this issue in our tests either since ForwardDiff is still a proper dependency there and always loaded.

PR #164, opened 2021-04-26
Load ForwardDiff correctly

devmotion:
This PR fixes https://github.com/TuringLang/DistributionsAD.jl/issues/163 for me locally.

@gideonsimpson can you check that it fixes the warning for you as well? You can install this branch with
```julia
julia> ] add DistributionsAD#dw/forwarddiff
```


gideonsimpson:
Yes, that cleaned it up for me.

Issue #168, opened 2021-05-04
MvCategorical not correct


bdeonovic:
It seems MvCategorical is not correct (unless I am doing something wrong)

```julia
julia> ps = rand(Dirichlet(ones(3)), 4);

julia> x = rand(1:3, 4);

julia> d1 = DistributionsAD.MvCategorical(ps);

julia> logpdf(d1,x)
-23.941191927580117

julia> d2 = Distributions.Product(Categorical.(eachcol(ps)));

julia> logpdf(d2,x)
-5.535670148277514
```

bdeonovic:
Hm, maybe I am doing something wrong

```julia
julia> sum(log.(vec(mapslices(prod, ps[x,1:4], dims=1))))
-23.941191927580114
```

bdeonovic:
```julia
Random.seed!(1234)
ps = rand(Dirichlet(ones(3)), 4);
x = rand(1:3, 4);
d1 = DistributionsAD.MvCategorical(ps);
logpdf(d1,x)

d2 = Distributions.Product(Categorical.(eachcol(ps)));
logpdf(d2,x)

#what DistributionsAD seems to be calculating
sum(log.(vec(mapslices(prod, ps[x,1:4], dims=1))))

#correct likelihood
log(prod([ps[x[i],i] for i in 1:length(x)]))
```

devmotion:
The problem is https://github.com/TuringLang/DistributionsAD.jl/blob/68d6748a103b4ee569dabce7fd82d937c9ebfbac/src/mvcategorical.jl#L159. Maybe this was the reason for adding MvCategorical in the first place? It won't work though since `view(ps, x, :)` just creates a matrix of values instead of selecting the `length(x)` appropriate values.

devmotion:
Yeah, apparently `logpdf` was one reason why it was added: https://github.com/TuringLang/DistributionsAD.jl/pull/29 :smile:

Issue #167, opened 2021-05-04
MvCategorical needs length method


bdeonovic:
```julia
julia> d = DistributionsAD.MvCategorical(rand(Dirichlet(ones(3)),4))
DistributionsAD.MvDiscreteNonParametric{Int64,Float64,Base.OneTo{Int64},Array{Float64,2}}(
support: Base.OneTo(3)
p: [0.05847351596221934 0.5434796327398905 0.6513029687383283 0.2455728666106314; 0.021949140453810377 0.35427920583290823 0.2725552514012437 0.7255803091385111; 0.9195773435839704 
0.10224116142720104 0.07614177986042824 0.028846824250857384]
)


julia> x = rand(1:3,4)
4-element Array{Int64,1}:
 2
 3
 2
 3

julia> logpdf(d,x)
ERROR: MethodError: no method matching length(::DistributionsAD.MvDiscreteNonParametric{Int64,Float64,Base.OneTo{Int64},Array{Float64,2}})
Closest candidates are:
  length(::Sampleable{Multivariate,S} where S<:ValueSupport) at C:\Users\yim583\.julia\packages\Distributions\cNe2C\src\common.jl:39
  length(::Sampleable) at C:\Users\yim583\.julia\packages\Distributions\cNe2C\src\common.jl:37
  length(::DataStructures.DiBitVector) at C:\Users\yim583\.julia\packages\DataStructures\ixwFs\src\dibit_vector.jl:40
  ...
Stacktrace:
 [1] length(::DistributionsAD.MvDiscreteNonParametric{Int64,Float64,Base.OneTo{Int64},Array{Float64,2}}) at C:\Users\yim583\.julia\packages\Distributions\cNe2C\src\common.jl:39     
 [2] logpdf(::DistributionsAD.MvDiscreteNonParametric{Int64,Float64,Base.OneTo{Int64},Array{Float64,2}}, ::Array{Int64,1}) at C:\Users\yim583\.julia\packages\Distributions\cNe2C\src\multivariates.jl:199
 [3] top-level scope at REPL[64]:1
```

devmotion:
[...]

PR #169, opened 2021-05-04
Remove `MvDiscreteNonParametric`

devmotion:
A quite drastic fix for https://github.com/TuringLang/DistributionsAD.jl/issues/168 and https://github.com/TuringLang/DistributionsAD.jl/issues/167.

The implementation is quite outdated, many upstream changes were not propagated, the log pdf is wrong, and AD seems to work nevertheless. I believe possible performance regressions should be fixed in a more targeted way and/or upstream.


mohamed82008:
I am not in favor of this. If it's wrong or out-dated, it should be fixed. If it should be upstreamed, then it should only be removed after the upstreaming is completed.

devmotion:
What is the reason for the custom implementation in DistributionsAD? If it was SubArray support or AD, both were fixed by upstream changes.

devmotion:
The current state is pretty bad since it is unclear why the implementation exists at all and almost everything is copied from a now outdated version of Distributions. This will be difficult to maintain in the future as well.

mohamed82008:
> What is the reason for the custom implementation in DistributionsAD?

Performance improvement for the vector case for AD.

devmotion:
Yeah but these performance improvements were achieved with an incorrect log pdf 😄

mohamed82008:
>Yeah but these performance improvements were achieved with an incorrect log pdf 😄

Possibly, but the solution is to fix it and check again.

PR #172, opened 2021-06-02
Update ChainRulesCore

devmotion:
In ChainRulesCore >= 0.9.44 some names were changed and deprecated, and then removed in ChainRulesCore 0.10. This PR updates ChainRulesCore and fixes these deprecations.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "DistributionsAD"
 uuid = "ced4e74d-a319-5a8a-b0ac-84af2272839c"
-version = "0.6.26"
+version = "0.6.27"
 
 [deps]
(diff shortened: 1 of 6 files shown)


devmotion:
Tests fail due to a problem in SpecialFunctions >= 1.4.1: https://github.com/JuliaMath/SpecialFunctions.jl/pull/325

devmotion:
For now I disabled the test that is broken by the upstream changes.

devmotion:
The remaining test error is interesting: it is caused by https://github.com/FluxML/Zygote.jl/blob/4d9b8d674f878ef8ef9db45ba4fee9d8aa94a73c/src/lib/array.jl#L66 which does not work for `T = ChainRulesCore.NotImplemented`.

Issue #158, opened 2021-03-30
TuringDirichlet returns non-zero pdf or error outside support


fernando-duarte:
(moving over from [#157](https://github.com/TuringLang/DistributionsAD.jl/issues/157#issuecomment-808339733) since it seems different enough)

The pdf of TuringDirichlet is not zero for points outside the support, and gives an error for points with negative coordinates. 

```Julia
using Pkg
Pkg.add(Pkg.PackageSpec(name="Distributions", version="0.24.15"))
Pkg.add(Pkg.PackageSpec(url="https://github.com/TuringLang/DistributionsAD.jl", rev="master"))
using Distributions, DistributionsAD

# test pdf of Dirichlet distribution
α = [0.43,0.21]
d = Distributions.Dirichlet(α)
Turing_d = DistributionsAD.TuringDirichlet(α)

x_not_sup = [0.5,0.8] # a point not in the support
x_neg = [-0.5,1.5] # a point not in the support with a negative entry

# Distributions.jl has the desired behavior
Distributions.pdf(d,x_not_sup) # returns 0.0
Distributions.pdf(d,x_neg)  # returns 0.0

# DistributionsAD.jl gives a different answer
DistributionsAD.pdf(Turing_d,x_not_sup) # a non-zero number
DistributionsAD.pdf(Turing_d,x_neg) # throws an error
```

Thanks!
Fernando

fernando-duarte:
Hi! Wondering if this issue made sense or if I should provide more info. 

Does it look somewhat easy to solve so that it makes sense that I wait for a fix? Or should I try to figure out a way around it assuming it is not fixable in a reasonable amount of time?

Thanks!

fernando-duarte:
Hi, wondering if there are any updates regarding a fix or if I should try to find another package or way to do it without DistributionsAD.jl.
Thank you.

yebai:
Thanks, @fernando-duarte for raising this issue.  @torfjelde Could you take a look at this issue, please?

yebai:
Fixed by #173. The fix will become available after https://github.com/JuliaRegistries/General/pull/38066 is merged.

fernando-duarte:
Thank you very much!

PR #173, opened 2021-06-02
Fix `simplex_logpdf` for values outside of the support

devmotion:
This addresses https://github.com/TuringLang/DistributionsAD.jl/issues/158 (at least partly, the differentials are not adjusted yet). I made these changes a while ago but forgot to open a PR.

In an ideal future we would just define the differentials (if needed) and remove the type and logpdf computation from DistributionsAD. But better we fix it as long as it exists :slightly_smiling_face:


devmotion:
The relevant tests pass, the error are caused by SpecialFunctions and would be fixed by https://github.com/JuliaMath/SpecialFunctions.jl/pull/325.

yebai (approved):
thanks, @devmotion - looks good to me!

PR #197, opened 2021-08-24
Fix warning, fix CI and run tests with Julia 1.3 only on Ubuntu

devmotion:


devmotion:
:eyes: :rofl: I don't know what's going on with the Zygote tests. Maybe we should just upgrade to ChainRulesCore 1 and see if they are fixed in more recent versions of Zygote.

DhairyaLGandhi:
Zygote moves to CR 1. This is a result of that. Perhaps we should not be too aggressive in adopting ChainRules.

devmotion:
>  Perhaps we should not be too aggressive in adopting ChainRules.

In Zygote? Or in DistributionsAD? In DistributionsAD it's only used for a handful of functions which still pass all tests. The errors seem to originate from the integration in Zygote, somehow now we have to define constructors for structs that worked automatically before?

DhairyaLGandhi:
I meant in Zygote. @oxinabox shouldn't these constructors be output as regular named tuples if CR doesn't have a compatible `rrule` and not wrap it in a `Tangent`?

DhairyaLGandhi:
Could we bump CR to 1, it's blocking https://github.com/FluxML/FluxBench.jl/pull/4 tests with the SciML stack

devmotion:
BTW locally I got the same errors with the latest release of Zygote and ChainRulesCore 1. Initially I only fixed the ChainRules definitions (which are not involved in the errors) but now I started to remove Zygote definitions that might not be needed anymore and ported some others. I still get the same errors regarding missing adjoints for constructors for e.g. product distributions of Bernoulli distributions.

devmotion:
> Could we bump CR to 1, it's blocking [FluxML/FluxBench.jl#4](https://github.com/FluxML/FluxBench.jl/pull/4) tests with the SciML stack

Zygote tests are completely broken also with CR 1 and the latest Zygote version, I would like to understand and fix the problem before making a new release. I don't mind if it uses CR1 and/or the latest Zygote version or not, I just want to fix the test failures right now.

oxinabox:
>  @oxinabox shouldn't these constructors be output as regular named tuples if CR doesn't have a compatible rrule and not wrap it in a Tangent?

I am going to need more context.
What constructors?

devmotion:
Of a large number of distributions: https://github.com/TuringLang/DistributionsAD.jl/pull/197/checks?check_run_id=3410352390

The univariate tests of these distributions seem to be fine though, only the multi- and matrixvariate product distributions seem to throw the missing adjoint error. So I assume it might be a problem with the `logpdf`/`loglikelihood` definitions for these cases, something that Zygote does not like about the definitions of the primals or tangents and somehow messes up the ChainRules integration.

devmotion:
I opened https://github.com/TuringLang/DistributionsAD.jl/pull/198 with updates for CR 1 but locally tests fail in a similar way.

oxinabox:
```
  Test threw exception
  Expression: ≈((Zygote.gradient(f, x))[1], finitediff, rtol = rtol, atol = atol)
  Need an adjoint for constructor TuringDirichlet{Float64, Vector{Float64}, Float64}. Gradient is of type Tangent{TuringDirichlet{Float64, Vector{Float64}, Float64}, NamedTuple{(:alpha, :alpha0, :lmnB), Tuple{Vector{Float64}, NoTangent, Float64}}}
 ```
 
Yeah that is weird, idk how they are getting there,
Those are not supposed to escape, being returned from a `rrule` into Zygote.
Working out how they are getting through seems needful

Adding https://github.com/JuliaDiff/ChainRules.jl/issues/510 might fix.
As might be just change https://github.com/FluxML/Zygote.jl/blob/975923915c6533abd278902aec6dfe07f1929f46/src/lib/lib.jl#L301
to actually accept them -- the data is correct it is just not expected to have a path that could get it out to here.

**Comment by devmotion** [comment]
Closed in favour of https://github.com/TuringLang/DistributionsAD.jl/pull/198.

**Comment by mcabbott** [comment]
The above example is fixed on latest Zygote, by https://github.com/FluxML/Zygote.jl/pull/1103 or the one before. 
```
julia> gradient(x -> sum(map(f, fill(A(x), 2, 1), ones(2, 1))), 2.0)
(2.0,)

(@v1.7) pkg> st Zygote
      Status `~/.julia/environments/v1.7/Project.toml`
  [e88e6eb3] Zygote v0.6.29
```

Issue #196, opened 2021-08-19
`BetaPrime` distribution has broken tests with the recent release of Distributions. 


yebai:
See https://github.com/TuringLang/DistributionsAD.jl/pull/185#issuecomment-901466657

devmotion:
It seems the choice of parameter alpha was just unfortunate since `alpha = 1` leads to the special case where `xlogy` ends up in the always-zero branch. Setting the parameter to 1.5, as done in the recent PRs, fixes the issue. Unfortunately, the PRs revealed many Zygote issues (somehow we end up with NamedTuple instead Tangent in some intermediate step and therefore Zygote throws an error) so the PRs can't be released currently.

devmotion:
More details: https://github.com/TuringLang/DistributionsAD.jl/pull/198#discussion_r699237854

PR #185, opened 2021-08-06
Fixes warnings during REPL startup while using custom SYSIMG's

bmharsha:
We encounter following warning during Julia REPL startup if we include DistributionsAD in the default SYSIMG, this PR fixes that issue

```
┌ Warning: Error requiring `ForwardDiff` from `DistributionsAD`
│   exception =
│    LoadError: SystemError: opening file "/home/bmharsha/.julia/packages/DistributionsAD/pj8zf/src/reversediffx.jl": No such file or directory
│    Stacktrace:
│      [1] systemerror(p::String, errno::Int32; extrainfo::Nothing)
│        @ Base ./error.jl:168
│      [2] #systemerror#62
│        @ ./error.jl:167 [inlined]
│      [3] systemerror
│        @ ./error.jl:167 [inlined]
│      [4] open(fname::String; lock::Bool, read::Nothing, write::Nothing, create::Nothing, truncate::Nothing, append::Nothing)
│        @ Base ./iostream.jl:293
│      [5] open
│        @ ./iostream.jl:282 [inlined]
│      [6] open(f::Base.var"#326#327"{String}, args::String; kwargs::Base.Iterators.Pairs{Union{}, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
│        @ Base ./io.jl:328
│      [7] open
│        @ ./io.jl:328 [inlined]
│      [8] read
│        @ ./io.jl:434 [inlined]
│      [9] _include(mapexpr::Function, mod::Module, _path::String)
│        @ Base ./loading.jl:1166
│     [10] include(mod::Module, _path::String)
│        @ Base ./Base.jl:386
│     [11] include(x::String)
│        @ DistributionsAD.ReverseDiffX ./reversediff.jl:1
│     [12] top-level scope
│        @ reversediff.jl:55
│     [13] eval
│        @ ./boot.jl:360 [inlined]
│     [14] include_string(mapexpr::typeof(identity), mod::Module, code::String, filename::String)
│        @ Base ./loading.jl:1116
│     [15] include_string(m::Module, txt::String, fname::String)
│        @ Base ./loading.jl:1126
│     [16] macro expansion
│        @ ~/.julia/packages/Requires/7Ncym/src/Requires.jl:42 [inlined]
│     [17] top-level scope
│        @ ~/.julia/packages/DistributionsAD/pj8zf/src/DistributionsAD.jl:74
│     [18] eval
│        @ ./boot.jl:360 [inlined]
│     [19] eval
[...]

PR #198, opened 2021-08-24
Fix test errors and update to ChainRules 1 and Distributions 0.25.15

devmotion:
This PR builds on https://github.com/TuringLang/DistributionsAD.jl/pull/197 and additionally updates the ChainRules definitions to ChainRules 1.

Unfortunately, locally Zygote tests seem to fail even with ChainRules 1. I opened a separate PR to not pollute https://github.com/TuringLang/DistributionsAD.jl/pull/197 with unrelated changes.

**Edit:**

This PR
- updates DistributionsAD to ChainRules 1
- removes ChainRules definitions of StatsFuns function which were moved to StatsFuns (https://github.com/JuliaStats/StatsFuns.jl/pull/106); StatsFuns dependency is bumped to 0.9.10 which contains these rules
- removes ChainRules definition of the pdf of `PoissonBinomial` which was moved to Distributions (https://github.com/JuliaStats/Distributions.jl/pull/1390); Distributions dependency is bumped to 0.25.15 which contains this rule
- removes code chunks for Distributions < 0.25
- fixes Zygote test errors (see below for more details)
- fixes https://github.com/TuringLang/DistributionsAD.jl/issues/196
- fixes https://github.com/TuringLang/DistributionsAD.jl/issues/199
- reduces CI: Julia 1.3 is tested only with Ubuntu
- fixes some deprecations of `MvNormal` constructors
- fixes https://github.com/TuringLang/DistributionsAD.jl/pull/195, https://github.com/TuringLang/DistributionsAD.jl/pull/194, https://github.com/TuringLang/DistributionsAD.jl/pull/193, https://github.com/TuringLang/DistributionsAD.jl/pull/192, https://github.com/TuringLang/DistributionsAD.jl/pull/190, https://github.com/TuringLang/DistributionsAD.jl/pull/189, https://github.com/TuringLang/DistributionsAD.jl/pull/188, https://github.com/TuringLang/DistributionsAD.jl/pull/186


devmotion:
Ah... NNlib with CR 1 only supports Julia 1.6. IIRC it's only used in the definition of a helper function in the tests so maybe we should just remove it from the test dependencies.

devmotion:
I added a workaround for the issue above, hopefully a similar workaround can fix the matrixvariate issues.

The StatsFuns CR definitions are removed since they were moved to StatsFuns (available in >= 0.9.10). I opened a PR to Distributions that transfers the CR definition for the pdf of `PoissonBinomial`: https://github.com/JuliaStats/Distributions.jl/pull/1390 If it is merged, we should remove the definitions in DistributionsAD.

test/runtests.jl
```diff
@@ -47,8 +45,23 @@ if GROUP == "All" || GROUP == "AD"
...
+    end
     to_simplex(x::AbstractArray{<:AbstractArray}) = to_simplex.(x)
+    function to_simplex_pullback(ȳ::AbstractArray, y::AbstractArray)
+        x̄ = ȳ .* y
+        x̄ .= x̄ .- y .* sum(x̄; dims=1)
+        return x̄
+    end
+    function ChainRulesCore.rrule(::typeof(to_simplex), x::AbstractArray{<:Real})
+        y = to_simplex(x)
+        pullback(ȳ) = (NoTangent(), to_simplex_pullback(ȳ, y))
```
oxinabox, line 62: May want to think about a `ProbectTo` here, but that is beyond scope of this PR I guess

oxinabox (approved):

devmotion:
Zygote tests pass again :tada:

oxinabox (approved):

src/zygote.jl
```diff
@@ -29,6 +29,24 @@ ZygoteRules.@adjoint function Distributions._logpdf(
     end
 end
 
+# Loglikelihood of multi- and matrixvariate distributions: multiple samples
+# workaround for Zygote issues discussed in
+# https://github.com/TuringLang/DistributionsAD.jl/pull/198
+ZygoteRules.@adjoint function Distributions.loglikelihood(
```
yebai, line 35: Thanks for looking into this - it looks good to me!

test/ad/distributions.jl
```diff
@@ -73,8 +73,8 @@
         DistSpec(Beta, (1.0, 2.0), 0.5),
 
         DistSpec(BetaPrime, (), 0.5),
-        DistSpec(BetaPrime, (1.0,), 0.5),
-        DistSpec(BetaPrime, (1.0, 2.0), 0.5),
+        DistSpec(BetaPrime, (1.5,), 0.5),
```
yebai, line 76: I also noticed the same issue but thought it only affects ReverseDiff with tape caching. When tape caching is disabled, ReverseDiff should handle control flows correctly (similar to Tracker).

yebai (approved):

yebai:
> @yebai I don't think we should move the ZygoteRules definitions in https://github.com/TuringLang/DistributionsAD.jl/blob/a603603e89fbafd4cfde6a69eb4c44874ca75cb7/src/zygote.jl to ChainRules. These are no useful differentiation rules but just hacks and workarounds for issues in Zygote 

It sounds good to keep them here, or depreciate them if possible, or move them into another package (e.g. `DynamicPPL`) where appropriate.

test/ad/distributions.jl
devmotion: I don't think it's ReverseDiff specific since e.g. Tracker failed as well (https://github.com/TuringLang/DistributionsAD.jl/pull/185/checks?check_run_id=3338494448#step:5:1097). In fact all AD systems apart from Zygote return a value that is different from finite differencing:
```julia
julia> using Distributions, FiniteDifferences, Tracker, ReverseDiff, ForwardDiff, Zygote

julia> f(x) = logpdf(BetaPrime(x), 0.5)

julia> central_fdm(5, 1)(f, 1.0)
0.495922603223565

julia> central_fdm(12, 1)(f, 1.0)
0.4959226032237455

julia> Tracker.gradient(f, 1.0)
(1.1890697837836721 (tracked),)

julia> ForwardDiff.derivative(f, 1.0)
1.1890697837836721

julia> ReverseDiff.gradient(f ∘ first, [1.0])
1-element Vector{Float64}:
 1.1890697837836721

julia> Zygote.gradient(f, 1.0)
(0.49592260322372683,)
```

The different values arise from the different behaviour of the AD systems for `x -> xlogy(x, 0.5)` at `0`. FiniteDifferences and Zygote return the correct derivative (`log(0.5)`) whereas Tracker, ForwardDiff, and ReverseDiff return 0:
```julia
julia> using LogExpFunctions

julia> central_fdm(12, 1)(x -> xlogy(x, 0.5), 0.0)
-0.6931471805599452

julia> Tracker.gradient(x -> xlogy(x, 0.5), 0.0)
(0.0 (tracked),)

julia> ForwardDiff.derivative(x -> xlogy(x, 0.5), 0.0)
0.0

julia> ReverseDiff.gradient(x -> xlogy(first(x), 0.5), [0.0])
1-element Vector{Float64}:
 0.0

julia> Zygote.gradient(x -> xlogy(x, 0.5), 0.0)
(-0.6931471805599453,)
```

It seems the AD systems handle the branch in https://github.com/JuliaStats/LogExpFunctions.jl/blob/39223ba1daa0244fae1885cd0bcba5a25743349b/src/basicfuns.jl#L29 differently/incorrectly.

PR #202, opened 2021-10-20
use `fill` adjoint from ChainRules

DhairyaLGandhi:
There might be reasons to overload it here, but the current implementation would override the generic `fill` adjoint, which can cause breakages. If you have an MWE, we can write a more targeted adjoint, and I 100% agree that our definitions should not assume numbers as eltypes of arrays. Presently, this causes https://github.com/SciML/NeuralPDE.jl/issues/412

cc @ChrisRackauckas

### DIFF
--- a/test/ad/utils.jl
+++ b/test/ad/utils.jl
@@ -81,6 +81,14 @@ function unpack_offset(x, offset, original::AbstractArray{<:AbstractArray})
     return val, newoffset
 end
 
+using DistributionsAD: ZygoteRules
+ZygoteRules.@adjoint function fill(x::Distribution, dims1::Int, dims2::Int...) where {T}
+    return ZygoteRules.pullback(x, dims1, dims2...) do x, dims1, dims2...
+        dims = (dims1, dims2...)
+        return reshape([x for i in 1:prod(dims)], dims)
+    end
+end
+
 # Run AD tests of a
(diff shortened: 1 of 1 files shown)


ChrisRackauckas (approved):
At least, this isn't where this definition should live.

devmotion:
So it seems some recent changes (last tests on master passed end of August with ReverseDiff and Tracker) in some upstream dependencies broke Tracker and ReverseDiff support. However, also Zygote errors, before tests were aborted mainly due to errors such as https://github.com/TuringLang/DistributionsAD.jl/pull/202/checks?check_run_id=3956438041#step:5:364 when testing arrays of distributions which potentially could be caused by the removal of the `fill` adjoint.

ChrisRackauckas:
>Then I guess a temporary workaround would be to just fix the dims::Tuple case and handle it correctly in the adjoint.

A better thing would be to narrow this dispatch. Which case is it actually fitting? Making it `Any` is clearly incorrect.

DhairyaLGandhi:
Moving it to tests would still define it in the tests, so it could make the DistributionsAD tests brittle indirectly. Either way, I want to see what CI says. I don't know if downstream testing is sufficient (better to do it than not, for sure).   @devmotion would you mind triggering CI

PR #203, opened 2021-10-21
Remove adjoint for `fill` and fix tests

devmotion:
This PR removes the Zygote adjoint for `fill` (see https://github.com/TuringLang/DistributionsAD.jl/pull/202 for a discussion for why it is bad and causes problems in other packages).

Additionally, it
- fixes some test errors of `Beta` and `Gamma` with ForwardDiff, Tracker, and ReverseDiff caused by `xlogy` by choosing less special parameter values (the proper upstream fix was https://github.com/JuliaDiff/DiffRules.jl/pull/69 but it has not been propagated to the AD packages yet)
- fixes deprecations and updates CI which was partly done in https://github.com/TuringLang/DistributionsAD.jl/pull/204 and reduces test times drastically
- improves the AD tests and in particular uses ChainRulesTestUtils for Zygote tests (Zygote tests pass now after https://github.com/FluxML/Zygote.jl/pull/1104 was merged; an integration test with DistributionsAD was added to Zygote which hopefully helps to avoid that Zygote breaks DistributionsAD accidentally in the future)

----

Original comment:

An alternative to https://github.com/TuringLang/DistributionsAD.jl/pull/202 with some additional unrelated fixes that make Zygote tests pass locally. Hopefully this fixes also the Tracker and ReverseDiff issues (didn't run them locally to save time).

Edit: Unfortunately, it seems the Tracker and ReverseDiff issues are not fixed by this PR.

Edit 2: Maybe the problem is that RD, Tracker, and FD don't know the derivative of `xlogy` and `xlog1py`. I'll check locally if adding some DiffRules helps.

Edit 3: Indeed, the issues with Tracker, ReverseDiff, and ForwardDiff are fixed by defining diff rules for `xlogy` and `xlog1py`. I made a PR to DiffRules (https://github.com/JuliaDiff/DiffRules.jl/pull/69) but unfortunately additionally it requires fixes for Tracker, ReverseDiff and ForwardDiff itself. In the mean time I guess the easiest is to choose different parameter values such that we don't end up with the special case for `xlogy` where `x` is zero.


test/ad/distributions.jl
```diff
@@ -471,10 +471,10 @@
             d.f(d.θ...) isa Union{VonMises,TriangularDist} && continue
 
             # Create `filldist` distribution
-            f_filldist = (θ...,) -> filldist(d.f(θ...), n...)
+            f_filldist = (θ...,) -> filldist(d.f(θ...), n1, n2)
 
             # Create `arraydist` distribution
-            f_arraydist = (θ...,) -> arraydist(fill(d.f(θ...), n...))
+            f_arraydist = (θ...,) -> arraydist([d.f(θ...) for _ in 1:n1, _ in 1:n2])
```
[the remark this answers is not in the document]
DhairyaLGandhi: This would error. It needs an adjoint for `Iterators.Product`. we can retain`fill(:: Distribution)`

test/ad/distributions.jl
```diff
@@ -471,10 +471,10 @@
             d.f(d.θ...) isa Union{VonMises,TriangularDist} && continue
 
             # Create `filldist` distribution
-            f_filldist = (θ...,) -> filldist(d.f(θ...), n...)
+            f_filldist = (θ...,) -> filldist(d.f(θ...), n1, n2)
```
devmotion, line 474:
```suggestion
            f_filldist = (θ...,) -> filldist(d.f(θ...), n...)
```

test/ad/distributions.jl
```diff
@@ -462,7 +462,7 @@
         end
 
         # Test `filldist` and `arraydist` distributions of univariate distributions
-        n = (2, 2) # always use 2 x 2 distributions
+        n = (n1, n2) = (2, 2) # always use 2 x 2 distributions
```
devmotion, line 465:
```suggestion
        n = (2, 2) # always use 2 x 2 distributions
```

test/ad/distributions.jl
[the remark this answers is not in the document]
DhairyaLGandhi: I'd do a quick mwe check and commit.

devmotion:
>     * evaluation of the loglikelihood of multiple samples from a product distribution of a matrix of uniform distributions errors:  https://github.com/TuringLang/DistributionsAD.jl/runs/3970857262?check_suite_focus=true#step:6:820 and similar lines below

@mcabbott I was able to fix these issues locally by simplifying the Zygote tests a bit (in contrast to the tracked/dual number type ADs we only have to check the full set of arguments once) and using `test_rrule` which takes care of vectorization for finite differencing etc. for us instead of the custom vectorization in the test suite. My plan is to change everything to CRTestutils and `test_rrule` and `test_frule`, and I already have a draft for ForwardDiff and Tracker as well, but originally my plan was to do it in a separate PR since this PR is quite messy already. However, it seems it might be the easiest way to fix one of the two remaining Zygote test issues.

However, I had to add the type piracy in https://github.com/TuringLang/DistributionsAD.jl/blob/12b9d34972108c35dffdec23f6c2c0240ea5b7c8/test/ad/utils.jl#L14-L26 (should be moved to Zygote) since otherwise some of the `filldist` tests failed (hopefully the last commit doesn't break other tests :smile:). More concretely, the problem was that https://github.com/JuliaDiff/ChainRulesTestUtils.jl/blob/dd0e246d1516451ec963db55a7a51910819082b3/src/testers.jl#L228 failed otherwise since cotangents that were specified as `NoTangent` and for which FiniteDifferences returned a `NoTangent` were computed as nested `Tangent` by Zygote. Unfortunately, I don't have a MWE but only a quite complicated test here.

devmotion:
I confirmed locally that the remaining Zygote test errors (namely, as mentioned above,
> evaluation of the loglikelihood of single and multiple samples from a product distribution of a vector of multivariate distributions errors

) are indeed fixed by https://github.com/FluxML/Zygote.jl/pull/1104. https://github.com/FluxML/Zygote.jl/pull/1104 also allows us to remove the workarounds for `Zygote.z2d` that are currently added to the tests.

@mcabbott Thus this PR is blocked by https://github.com/FluxML/Zygote.jl/pull/1104 but tests should pass once a new Zygote version with the changes in this PR is released.

test/ad/utils.jl
```diff
@@ -1,6 +1,156 @@
...
+        deltas = map(n -> get(delta, n, nothing), fnames)
+        primals = map(n -> getfield(primal, n), fnames)
+        inner = map(Zygote.z2d, deltas, primals)
+        return if inner isa Tuple{Vararg{AbstractZero}}
+            NoTangent()
+        else
+            backing = NamedTuple{fnames}(inner)
+            canonicalize(Tangent{T, typeof(backing)}(backing))
+        end
+    end
```
devmotion, line 29: The Zygote PR was merged and released:
```suggestion
```

yebai (approved):
Looks good to me - many thanks @devmotion !

PR #204, opened 2021-10-21
Extension of #203: Fix deprecations in test

devmotion:
This PR fixes some deprecations in the tests (only non-deprecated calls are tested) which hopefully also decrease testing times. I kept it separate from #203 to make it easier to review.


devmotion:
This is a major improvement: the ReverseDiff test was completed and failed after 12 minutes whereas in #203 it was still running after 45 minutes.

PR #207, opened 2021-11-25
Fix `_wishart_genA!`

devmotion:
Broken by https://github.com/JuliaStats/Distributions.jl/pull/1429.

(I really hate these reimplementations that are based on outdated and diverging copies of code from Distributions...)


yebai (approved):
yeah, I'm hoping that most the code here will go into `Distributions` or other source packages in the longer term.

yebai (approved):

PR #208, opened 2021-11-26
Do not use internal functions from Distributions

devmotion:
This is required for https://github.com/JuliaStats/Distributions.jl/pull/1425.


yebai (approved):
looks good - thanks @devmotion!

PR #209, opened 2022-01-05
reexport Distributions

mileslucas:
- add Reexport for Distributions
- implicitly test reexport by removing 'using Distributions'
- bump project version to 0.6.36


devmotion:
I think it is good that DistributionsAD does not re-export Distributions, and therefore in my opinion this PR should not be merged:

- DistributionsAD is a workaround for AD issues of Distributions that should rather be fixed in Distributions or other upstream packages. It is full of type piracy and the long term goal is to makr it obsolete (some things are already fixed upstream, and for some things upstream PRs exist). As such, typically users and packages should not just load DistributionsAD but rather Distributions and only load DistributionsAD explicitly if they face problems and are aware of the problems with DistributionsAD. Hence re-exporting Distributions would encourage non-desired code patterns.
- More generally, re-exporting is problematic: If we want to support a new breaking release of Distributions we have to tag a breaking release as well, even if the implementation in DistributionsAD itself is not affected by the breaking change upstream.

mileslucas:
Those are good points, I don't have any rebuttals.

If there aren't any outstanding reason to counter your arguments, I'll close this in a day or two.

PR #211, opened 2022-01-22
Remove `TuringUniform`

devmotion:
Now that https://github.com/JuliaStats/Distributions.jl/pull/1459 is released we can remove `TuringUniform`, its rules and the Zygote workaround for `Uniform`.


yebai (approved):
thanks @devmotion - looks good!

PR #212, opened 2022-01-25
Remove pullback for `Wishart`

devmotion:
This should be fixed by https://github.com/JuliaStats/Distributions.jl/pull/1410.

### DIFF
--- a/src/zygote.jl
+++ b/src/zygote.jl
@@ -35,30 +35,6 @@ end
 
 ## Wishart ##
 
-# Custom adjoint since Zygote can't differentiate through `@warn`
-# TODO: Remove when fixed upstream in Distributions
-ZygoteRules.@adjoint function Wishart(df::T, S::AbstractPDMat{T}, warn::Bool = true) where T<:Real
-    function _Wishart(df::T, S::AbstractPDMat{T}, warn::Bool = true) where T
-        df > 0 || throw(ArgumentError("df must be positive. got $(df)."))
-        p = dim(S)
-        rnk = p
-        singular = df <= p - 1
-        if singular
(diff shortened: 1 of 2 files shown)


src/zygote.jl
```diff
@@ -35,30 +35,6 @@ end
...
-        prom_S = convert(AbstractArray{T}, S)
-        Wishart{R, typeof(prom_S), typeof(rnk)}(R(df), prom_S, R(logc0), rnk, singular)
-    end
-    return ZygoteRules.pullback(_Wishart, df, S, warn)
-end
-
-_warn(msg) = @warn(msg)
-@non_differentiable _warn(msg)
-
 ZygoteRules.@adjoint function Distributions.Wishart(df::Real, S::AbstractMatrix{<:Real})
```
yebai, line 38: Do we still need to keep this `adjoined?

yebai (approved):
Thanks, @devmotion - it looks good other than the minor comment below.

src/zygote.jl
devmotion: I wasn't sure and hence planned to check it separately. I guess its motivation is that it avoids the construction of a `PDMat` and instead works with the Cholesky factors directly (https://github.com/TuringLang/DistributionsAD.jl/blob/9461c6b9eb3472efce160d37fbe1a1b4ec8c9bb6/src/matrixvariate.jl#L7-L11).

devmotion: Is it fine with you if we merge this PR and investigate the question later, together with other PDMats workarounds @yebai?

yebai: No problem - happy to merge this PR as-is!

PR #215, opened 2022-02-17
Drop support for Julia < 1.6

devmotion:
ChainRulesCore broke DistributionsAD recently. We noticed this in the integration tests in Distributions and it was fixed in a new release of ChainRules (that's a bit weird but seemed to follow the general policy: https://github.com/JuliaDiff/ChainRules.jl/pull/571#issuecomment-1025186952). However, ChainRulesCore (with the problematic commit) still supports Julia 1.0 whereas the fix in ChainRules is only available in Julia 1.6. Maybe this should be fixed upstream (since it also affects previous releases of DistributionsAD) but for future versions we can avoid the problem by dropping support for Julia < 1.6.


devmotion:
There's some discussion about how to fix this issue upstream, so maybe this PR is not needed.

devmotion:
Fixed upstream.

PR #227, opened 2022-07-06
Faster filldist()

alyst:
Switch `logpdf(::FillDist)` from `sum(map(f, x))` to `mapreduce(f, +, x)` to eliminate the unnecessary array allocations.

`sum(f, x)` would have been even simpler, but I'm hitting `*(::TrackedReal, ::Dual)` method ambiguity on this path.

It's beyond my current understanding of the autodiff code to figure out whether just this change breaks some tests (`Chernoff` distribution) and fixes other (`Zygote`) or it is just a coincidence.

I have tried to `@benchmark` the change, but with `@benchmark(logpdf($(filldist(Cauchy(), 1000)), $(rand(1000))))` I don't see much of the change.
I guess it starts to be visible with dual numbers.

So here's the profiling results.

Before:

After:

Note that array allocation and copying disappear, and it looks like overall it should be ~30% faster.

### DIFF
--- a/src/arraydist.jl
+++ b/src/arraydist.jl
@@ -22,10 +22,10 @@ function Distributions._logpdf(dist::MatrixOfUnivariate, x::AbstractMatrix{<:Rea
     return sum(Broadcast.instantiate(Broadcast.broadcasted(logpdf, dist.dists, x)))
 end
 function Distributions.logpdf(dist::MatrixOfUnivariate, x::AbstractArray{<:AbstractMatrix{<:Real}})
-    return map(x -> logpdf(dist, x), x)
+    return map(Base.Fix1(logpdf, dist), x)
 end
 function Distributions.logpdf(dist::MatrixOfUnivariate, x::AbstractArray{<:Matrix{<:Real}})
-    return map(x -> logpdf(dist, x), x)
+    return map(Base.Fix1(logpdf, dist), x)
 end
 
 function Distributions.rand(rng::Random.AbstractRNG, dist::MatrixOfUnivariate)
@@ -51,13 +51,13 @@ function Distributions._logpdf(dist::VectorOfMultivariate, x::AbstractMatrix{<:R
     return sum(((di, xi),) -> logpdf(di, xi), zip(dist.dists, eachcol(x)))
 end
 function Distributions.logpdf(dist::VectorOfMultivariate, x::AbstractArray{<:AbstractMatrix{<:Real}})
-    return map(x -> logpdf(dist, x), x)
+    return map(Base.Fix1(logpdf, dist), x)
 end
 function Distributions.logpdf(dist::VectorOfMultivariate, x::AbstractArray{<:Matrix{<:Real}})
-    return map(x -> logpdf(dist, x), x)
+    return map(Base.Fix1(logpdf, dist), x)
 end
 
 function Distributions.rand(rng::Random.AbstractRNG, dist::VectorOfMultivariate)
     init = reshape(rand(rng, dist.dists[1]), :, 1)
-    return mapreduce(i -> rand(rng, dist.dists[i]), hcat, 2:length(dist); init = init)
+    return mapreduce(Base.Fix1(rand, rng), hcat, view(dist.dists, 2:length(dist)); init = init)
 end
--- a/src/arraydist.jl
+++ b/src/arraydist.jl
@@ -3,7 +3,7 @@
 const VectorOfUnivariate = Distributions.Product
 
 function arraydist(dists::AbstractVector{<:UnivariateDistribution})
-    return Product(dists)
+    return product_distribution(dists)
 end
 
 struct MatrixOfUnivariate{
--- a/src/arraydist.jl
+++ b/src/arraydist.jl
@@ -18,7 +18,9 @@ function arraydist(dists::AbstractMatrix{<:UnivariateDistribution})
     return MatrixOfUnivariate(dists)
 end
 function Distributions._logpdf(dist::MatrixOfUnivariate, x::AbstractMatrix{<:Real})
-    return mapreduce(logpdf, +, dist.dists, x)
+    # Lazy broadcast to avoid allocations and use pairwise summation
+    y = Broadcast.instantiate(Broadcast.broadcasted(logpdf, dist.dists, x))
+    return sum(y)
 end
 function Distributions.logpdf(dist::MatrixOfUnivariate, x::AbstractArray{<:AbstractMatrix{<:Real}})
     return map(x -> logpdf(dist, x), x)
--- a/src/arraydist.jl
+++ b/src/arraydist.jl
@@ -18,7 +18,7 @@ function arraydist(dists::AbstractMatrix{<:UnivariateDistribution})
     return MatrixOfUnivariate(dists)
 end
 function Distributions._logpdf(dist::MatrixOfUnivariate, x::AbstractMatrix{<:Real})
-    return mapreduce((disti, xi) -> logpdf(disti, xi), +, dist.dists, x)
+    return mapreduce(logpdf, +, dist.dists, x)
 end
 function Distributions.logpdf(dist::MatrixOfUnivariate, x::AbstractArray{<:AbstractMatrix{<:Real}})
     return map(x -> logpdf(dist, x), x)
--- a/src/arraydist.jl
+++ b/src/arraydist.jl
@@ -3,7 +3,10 @@
 const VectorOfUnivariate = Distributions.Product
 
 function arraydist(dists::AbstractVector{<:UnivariateDistribution})
-    return product_distribution(dists)
+    V = typeof(dists)
+    T = eltype(dists)
+    S = Distributions.value_support(T)
+    return Product{S,T,V}(dists)
 end
 
 struct MatrixOfUnivariate{
--- a/src/arraydist.jl
+++ b/src/arraydist.jl
@@ -48,7 +48,7 @@ function arraydist(dists::AbstractVector{<:MultivariateDistribution})
 end
 
 function Distributions._logpdf(dist::VectorOfMultivariate, x::AbstractMatrix{<:Real})
-    return sum(((di, xi),) -> logpdf(di, xi), zip(dist.dists, eachcol(x)))
+    return sum(Broadcast.instantiate(Broadcast.broadcasted(logpdf, dist.dists, eachcol(x))))
 end
 function Distributions.logpdf(dist::VectorOfMultivariate, x::AbstractArray{<:AbstractMatrix{<:Real}})
     return map(Base.Fix1(logpdf, dist), x)
--- a/src/arraydist.jl
+++ b/src/arraydist.jl
@@ -19,8 +19,7 @@ function arraydist(dists::AbstractMatrix{<:UnivariateDistribution})
 end
 function Distributions._logpdf(dist::MatrixOfUnivariate, x::AbstractMatrix{<:Real})
     # Lazy broadcast to avoid allocations and use pairwise summation
-    y = Broadcast.instantiate(Broadcast.broadcasted(logpdf, dist.dists, x))
-    return sum(y)
+    return sum(Broadcast.instantiate(Broadcast.broadcasted(logpdf, dist.dists, x)))
 end
 function Distributions.logpdf(dist::MatrixOfUnivariate, x::AbstractArray{<:AbstractMatrix{<:Real}})
     return map(x -> logpdf(dist, x), x)
--- a/src/arraydist.jl
+++ b/src/arraydist.jl
@@ -18,9 +18,7 @@ function arraydist(dists::AbstractMatrix{<:UnivariateDistribution})
     return MatrixOfUnivariate(dists)
 end
 function Distributions._logpdf(dist::MatrixOfUnivariate, x::AbstractMatrix{<:Real})
-    # return sum(((d, xi),) -> logpdf(d, xi), zip(dist.dists, x))
-    # Broadcasting here breaks Tracker for some reason
-    return sum(map(logpdf, dist.dists, x))
+    return mapreduce((disti, xi) -> logpdf(disti, xi), +, dist.dists, x)
 end
 function Distributions.logpdf(dist::MatrixOfUnivariate, x::AbstractArray{<:AbstractMatrix{<:Real}})
     return map(x -> logpdf(dist, x), x)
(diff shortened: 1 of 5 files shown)


src/arraydist.jl
```diff
@@ -18,9 +18,7 @@ function arraydist(dists::AbstractMatrix{<:UnivariateDistribution})
     return MatrixOfUnivariate(dists)
 end
 function Distributions._logpdf(dist::MatrixOfUnivariate, x::AbstractMatrix{<:Real})
-    # return sum(((d, xi),) -> logpdf(d, xi), zip(dist.dists, x))
-    # Broadcasting here breaks Tracker for some reason
-    return sum(map(logpdf, dist.dists, x))
+    return mapreduce((disti, xi) -> logpdf(disti, xi), +, dist.dists, x)
```
devmotion, line 21: Note that `mapreduce` does not check that `dist.dists` and `x` are of the same size. But that should be fine since checks should happen before `_logpdf` is called.
```suggestion
    return mapreduce(logpdf, +, dist.dists, x)
```

src/filldist.jl
```diff
@@ -30,21 +30,18 @@ end
 function _flat_logpdf(dist, x)
     if toflatten(dist)
         f, args = flatten(dist)
-        return sum(f.(args..., x))
+        return mapreduce(xi -> f(args..., xi), +, x)
     else
-        return sum(map(x) do x
-            logpdf(dist, x)
-        end)
+        return mapreduce(xi -> logpdf(dist, xi), +, x)
```
devmotion, line 35:
```suggestion
        return mapreduce(Base.Fix1(logpdf, dist), +, x)
```

src/filldist.jl
```diff
@@ -30,21 +30,18 @@ end
...
 
 function _flat_logpdf_mat(dist, x)
     if toflatten(dist)
         f, args = flatten(dist)
-        return vec(sum(f.(args..., x), dims = 1))
+        return vec(mapreduce(xi -> f(args..., xi), +, x, dims = 1))
     else
-        temp = map(x -> logpdf(dist, x), x)
-        return vec(sum(temp, dims = 1))
+        return vec(mapreduce(xi -> logpdf(dist, xi), +, x, dims = 1))
```
devmotion, line 44:
```suggestion
        return vec(mapreduce(Base.Fix1(logpdf, dist), +, x; dims = 1))
```

devmotion (commented):
Thanks for the PR. I guess these methods are mainly optimized for compatibility with most AD backends (and in particular Zygote) and hence for historical reasons used whatever was possible and inferred at the time when they were introduced.

(A side remark: The long term plan is to remove all of this functionality from DistributionsAD and generalize it in Distributions: https://github.com/JuliaStats/Distributions.jl/pull/1391)

alyst:
Is there any chance to merge it?

devmotion:
Sure, if tests pass successfully I assume it can be approved by a maintainer and merged eventually.

alyst:
> Sure, if tests pass successfully I assume it can be approved by a maintainer and merged eventually.

I hope the fact that the tests got cancelled for some backends is some temporary issue that is not related to this PR. Otherwise it would be hard to know whether the tests pass.

src/filldist.jl
```diff
@@ -30,21 +30,18 @@ end
...
-        end)
+        return mapreduce(Base.Fix1(logpdf, dist), +, x)
     end
 end
 
 function _flat_logpdf_mat(dist, x)
     if toflatten(dist)
         f, args = flatten(dist)
-        return vec(sum(f.(args..., x), dims = 1))
+        return vec(mapreduce(xi -> f(args..., xi), +, x, dims = 1))
```
devmotion, line 42:
```suggestion
        return vec(sum(xi -> f(args..., xi), x; dims=1))
```

src/filldist.jl
```diff
@@ -30,21 +30,18 @@ end
...
     else
-        temp = map(x -> logpdf(dist, x), x)
-        return vec(sum(temp, dims = 1))
+        return vec(mapreduce(Base.Fix1(logpdf, dist), +, x; dims = 1))
```
devmotion, line 44:
```suggestion
        return vec(sum(Base.Fix1(logpdf, dist), x; dims = 1))
```

test/ad/distributions.jl
```diff
@@ -409,12 +409,11 @@
...
             # https://github.com/TuringLang/DistributionsAD.jl/pull/172#issuecomment-853721493
             filldist_broken = if D <: Skellam
-                ((d.broken..., :Zygote, :ReverseDiff), (d.broken..., :Zygote, :ReverseDiff))
+                ((d.broken..., :ReverseDiff), (d.broken..., :ReverseDiff))
             elseif D <: PoissonBinomial
                 ((d.broken..., :Zygote), (d.broken..., :Zygote))
             elseif D <: Chernoff
-                # Zygote is not broken with `filldist`
-                ((), ())
+                ((d.broken..., :Zygote), (d.broken..., :Zygote))
```
devmotion, line 416: `Chernoff` is already marked as broken for `:Zygote`. The branch explicitly removed this since `filldist` was not broken with Zygote. But if this PR breaks `filldist` with `Zygote` you can just remove the whole branch.
```suggestion
```

src/filldist.jl
```diff
@@ -30,21 +30,18 @@ end
 function _flat_logpdf(dist, x)
     if toflatten(dist)
         f, args = flatten(dist)
-        return sum(f.(args..., x))
+        return mapreduce(xi -> f(args..., xi), +, x)
```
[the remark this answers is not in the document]
alyst: Let's try, I hope it would not break Tracker.

src/arraydist.jl
```diff
@@ -18,9 +18,7 @@ function arraydist(dists::AbstractMatrix{<:UnivariateDistribution})
     return MatrixOfUnivariate(dists)
 end
 function Distributions._logpdf(dist::MatrixOfUnivariate, x::AbstractMatrix{<:Real})
-    # return sum(((d, xi),) -> logpdf(d, xi), zip(dist.dists, x))
-    # Broadcasting here breaks Tracker for some reason
-    return sum(map(logpdf, dist.dists, x))
+    return mapreduce(logpdf, +, dist.dists, x)
```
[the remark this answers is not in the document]
alyst: > Are you sure this is helpful? At least it should still allocate since it just calls `reduce(+, map(logpdf, dist.dists, x))` So it seems to be the same as before but less obvious?

It guess you are right. Well, this is quite unfortunate, since the doc of `mapreduce()` explicitly states that it is there to avoid allocation.
Calling low-level broadcast routines each time you want `mapreduce()` looks like a design deficiency on the core Julia side.
There is JuliaLang/julia#41001 that should address it.
But meanwhile I'll just apply your suggestion.

src/filldist.jl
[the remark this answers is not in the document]
alyst: Alas, this triggers `Tracker` method ambiguity
```
Multivariate distributions: Error During Test at C:\Users\astukalov\.julia\dev\DistributionsAD\test\ad\utils.jl:357
  Test threw exception
  Expression: ≈(Tracker.data((Tracker.gradient(f, x))[1]), finitediff, rtol = rtol, atol = atol)
  MethodError: -(::ForwardDiff.Dual{Nothing, Float64, 1}, ::Tracker.TrackedReal{Float64}) is ambiguous. Candidates:
    -(a::Real, b::Tracker.TrackedReal) in Tracker at C:\Users\astukalov\.julia\packages\Tracker\9xWLl\src\lib\real.jl:96
    -(x::ForwardDiff.Dual{Tx}, y::Real) where Tx in ForwardDiff at C:\Users\astukalov\.julia\packages\ForwardDiff\wAaVJ\src\dual.jl:144
  Possible fix, define
    -(::ForwardDiff.Dual{Tx}, ::Tracker.TrackedReal) where Tx
  Stacktrace:
    [1] zval(μ::Tracker.TrackedReal{Float64}, σ::Tracker.TrackedReal{Float64}, x::ForwardDiff.Dual{Nothing, Float64, 1})
      @ StatsFuns C:\Users\astukalov\.julia\packages\StatsFuns\vxSkw\src\distrs\norm.jl:10
    [2] normlogpdf(μ::Tracker.TrackedReal{Float64}, σ::Tracker.TrackedReal{Float64}, x::ForwardDiff.Dual{Nothing, Float64, 1})
      @ StatsFuns C:\Users\astukalov\.julia\packages\StatsFuns\vxSkw\src\distrs\norm.jl:39
    [3] logpdf
      @ C:\Users\astukalov\.julia\packages\Distributions\QLJcf\src\univariate\continuous\logitnormal.jl:126 [inlined]
    [4] #200
      @ .\none:0 [inlined]
    [5] #21
      @ C:\Users\astukalov\.julia\dev\DistributionsAD\src\filldist.jl:33 [inlined]
    [6] partial(f::DistributionsAD.var"#21#22"{Tuple{Tracker.TrackedReal{Float64}, Tracker.TrackedReal{Float64}}, DistributionsAD.var"#200#201"}, Δ::Tracker.TrackedReal{Float64}, i::Int64, args::Float64)
      @ Tracker C:\Users\astukalov\.julia\packages\Tracker\9xWLl\src\lib\array.jl:573
    [7] _broadcast_getindex_evalf
      @ .\broadcast.jl:670 [inlined]
    [8] _broadcast_getindex
      @ .\broadcast.jl:643 [inlined]
    [9] getindex
      @ .\broadcast.jl:597 [inlined]
   [10] copy
      @ .\broadcast.jl:899 [inlined]
   [11] materialize(bc::Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{1}, Nothing, typeof(Tracker.partial), Tuple{Base.RefValue{DistributionsAD.var"#21#22"{Tuple{Tracker.TrackedReal{Float64}, Tracker.TrackedReal{Float64}}, DistributionsAD.var"#200#201"}}, Vector{Tracker.TrackedReal{Float64}}, Int64, Vector{Float64}}})
      @ Base.Broadcast .\broadcast.jl:860
   [12] broadcast(::typeof(Tracker.partial), ::Base.RefValue{DistributionsAD.var"#21#22"{Tuple{Tracker.TrackedReal{Float64}, Tracker.TrackedReal{Float64}}, DistributionsAD.var"#200#201"}}, ::Vector{Tracker.TrackedReal{Float64}}, ::Int64, ::Vararg{Any})
      @ Base.Broadcast .\broadcast.jl:798
   [13] ∇broadcast
      @ C:\Users\astukalov\.julia\packages\Tracker\9xWLl\src\lib\array.jl:577 [inlined]
   [14] copy(bc::Base.Broadcast.Broadcasted{Tracker.TrackedStyle, Tuple{Base.OneTo{Int64}}, typeof(Tracker.partial), Tuple{Base.RefValue{DistributionsAD.var"#21#22"{Tuple{Tracker.TrackedReal{Float64}, Tracker.TrackedReal{Float64}}, DistributionsAD.var"#200#201"}}, Vector{Tracker.TrackedReal{Float64}}, Int64, TrackedArray{…,Vector{Float64}}}})
      @ Tracker C:\Users\astukalov\.julia\packages\Tracker\9xWLl\src\lib\array.jl:607
   [15] materialize
      @ .\broadcast.jl:860 [inlined]
   [16] (::Tracker.var"#621#623"{Vector{Tracker.TrackedReal{Float64}}, DistributionsAD.var"#21#22"{Tuple{Tracker.TrackedReal{Float64}, Tracker.TrackedReal{Float64}}, DistributionsAD.var"#200#201"}, Tuple{TrackedArray{…,Vector{Float64}}}})(i::Int64)
      @ Tracker C:\Users\astukalov\.julia\packages\Tracker\9xWLl\src\lib\array.jl:580
```
I guess it has to be fixed on the `Tracker` side. Also looks like it still generates an array.

alyst:
This is my best attempt to make the tests pass. They take quite long (that is why CI times out), and actually don't complete. Here's the message I'm getting:
```
[ Info: Testing: arraydist(MvNormal, 2)
[ Info: Testing: filldist(MvNormal, 2, 2)
[ Info: Testing: arraydist(MvNormal, 2, 2)
ERROR: Package DistributionsAD errored during testing (exit code: 3221225727)
```
I'm not sure whether it is related to the PR though.
Otherwise there tests pass except for `Chernoff` distribution, which I don't know how to fix, because it looks like some tests for Chernoff distribution pass and some don't.
(And, frankly, the current design of enabling/disabling specific tests is not very straightforward for the newcomer to digest)

src/arraydist.jl
```diff
@@ -3,7 +3,7 @@
 const VectorOfUnivariate = Distributions.Product
 
 function arraydist(dists::AbstractVector{<:UnivariateDistribution})
-    return Product(dists)
+    return product_distribution(dists)
```
devmotion, line 6: This changes the return type and breaks tests, as observed in https://github.com/TuringLang/DistributionsAD.jl/pull/228#discussion_r928239304.

yebai, line 6:
```suggestion
    V = typeof(dists)
    T = eltype(dists)
    S = Distributions.value_support(T)
    return Product{S,T,V}(dists)
```

yebai: Let's see whether #228 fixes broken tests.

test/ad/distributions.jl
```diff
@@ -408,13 +408,10 @@
...
             # https://github.com/TuringLang/DistributionsAD.jl/pull/172#issuecomment-853721493
-            filldist_broken = if D <: Skellam
-                ((d.broken..., :Zygote, :ReverseDiff), (d.broken..., :Zygote, :ReverseDiff))
-            elseif D <: PoissonBinomial
+            filldist_broken = if D <: PoissonBinomial
                 ((d.broken..., :Zygote), (d.broken..., :Zygote))
             elseif D <: Chernoff
-                # Zygote is not broken with `filldist`
-                ((), ())
+                ((d.broken..., :Zygote), (d.broken..., :Zygote))
```
devmotion, line 414: Seems it is not broken (as before), see https://github.com/TuringLang/DistributionsAD.jl/runs/7724467303?check_suite_focus=true#step:6:420:
```suggestion
                # Zygote is not broken with `filldist`
                ((), ())
```

src/filldist.jl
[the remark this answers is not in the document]
devmotion: > I guess it has to be fixed on the `Tracker` side. Also looks like it still generates an array.

Well, Tracker defines `sum(f, xs::TrackedArray)` as `sum(f.(xs))`: https://github.com/FluxML/Tracker.jl/blob/84ff74daaa250dadb424d251c66d64fa64ade819/src/lib/array.jl#L358 So it will generate an intermediate array and will use ForwardDiff when differentiating (as broadcasting operations with Tracker involve ForwardDiff).

PR #228, opened 2022-07-24
Fix deprecation warnings of `Product` constructor

torfjelde:
`Product` is deprecated in favour of `product_distribution` in Distributions >0.25.64.

### DIFF
--- a/src/arraydist.jl
+++ b/src/arraydist.jl
@@ -3,7 +3,7 @@
 const VectorOfUnivariate = Distributions.Product
 
 function arraydist(dists::AbstractVector{<:UnivariateDistribution})
-    return Product(dists)
+    return product_distribution(dists)
 end
 
 struct MatrixOfUnivariate{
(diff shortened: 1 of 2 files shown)


src/arraydist.jl
```diff
@@ -3,7 +3,7 @@
 const VectorOfUnivariate = Distributions.Product
 
 function arraydist(dists::AbstractVector{<:UnivariateDistribution})
-    return Product(dists)
+    return product_distribution(dists)
```
devmotion, line 6: That's the right thing to do since it will create optimized product distributions whenever possible - but apparently the optimization for `Normal` (creating a `MvNormal`) breaks Tracker. So I think either we have to add an optimization for `product_distribution(::AbstractVector{Normal{<:Tracker.TrackedReal}})` (type piracy), improve the definition in Distributions.jl, or use something like

```suggestion
    V = typeof(dists)
    T = eltype(dists)
    S = value_support(T)
    return Product{V,T,S}(dists)
```

I tend towards the last option as a temporary workaround since it doesn't change the behaviour and resulting types. Unfortunately, it misses the optimizations and maybe (?) requires us to bump the Distributions compat (the inner constructor was only added explicitly in 0.25.66 IIRC).

ChrisRackauckas:
Bump?

yebai (approved):

yebai:
Many thanks, @torfjelde and @devmotion!

PR #229, opened 2022-09-25
Fix deprecations

devmotion:
This PR should fix the deprecation warnings (deprecated `Val(false)` argument in `cholesky` and deprecated `LocationScale`).


test/ad/distributions.jl
```diff
@@ -143,7 +143,7 @@
         DistSpec(Levy, (0.0,), 0.5),
         DistSpec(Levy, (0.0, 2.0), 0.5),
 
-        DistSpec((a, b) -> LocationScale(a, b, Normal()), (1.0, 2.0), 0.5),
+        DistSpec((a, b) -> a + b * Beta(), (1.0, 2.0), 0.5),
```
devmotion, line 146:
```suggestion
        DistSpec((a, b) -> a + b * Beta(), (1.0, 2.0), 2.0),
```

test/ad/distributions.jl
```diff
@@ -143,7 +143,7 @@
         DistSpec(Levy, (0.0,), 0.5),
         DistSpec(Levy, (0.0, 2.0), 0.5),
 
-        DistSpec((a, b) -> LocationScale(a, b, Normal()), (1.0, 2.0), 0.5),
+        DistSpec((a, b) -> a + b * Beta(), (1.0, 2.0), 2.0),
```
yebai, line 146: Maybe consider adding a comment here to indicate it is meant to test an `AffineDistribution` distribution. The syntax `a+b*Beta()` is quite subtle, I think.

yebai (approved):
Many thanks, @devmotion - looks good except for one minor clarity comment below.

test/ad/distributions.jl
devmotion:
```suggestion
        # Test AffineDistribution
        DistSpec((a, b) -> a + b * Beta(), (1.0, 2.0), 2.0),
```

PR #231, opened 2023-01-16
Faster `arraydist` with LazyArrays.jl

torfjelde:
This PR is basically an accumulation of the discussion in https://github.com/TuringLang/Turing.jl/issues/1934 and https://github.com/TuringLang/DistributionsAD.jl/pull/230.

It's a hack to make reverse-mode AD packages that uses ForwardDiff for broadcasting _much_ faster when used in combination with LazyArrays.jl.

Unfortunately, this requires a rather ugly hack that is `make_closure` (maybe there's a more elegant solution? @devmotion pls halp!), _but_ it does buy us a whole lot of runtime.

### DIFF
--- a/src/DistributionsAD.jl
+++ b/src/DistributionsAD.jl
@@ -80,45 +80,7 @@ include("zygote.jl")
     end
 
     @require LazyArrays = "5078a376-72f3-5289-bfd5-ec5146d43c02" begin
-        using .LazyArrays: BroadcastArray, BroadcastVector, LazyArray
-
-        const LazyVectorOfUnivariate{
-            S<:ValueSupport,
-            T<:UnivariateDistribution{S},
-            Tdists<:BroadcastVector{T},
-        } = VectorOfUnivariate{S,T,Tdists}
-
-        function Distributions._logpdf(
-            dist::LazyVectorOfUnivariate,
-            x::AbstractVector{<:Real},
-        )
-            return sum(copy(logpdf.(dist.v, x)))
-        end
-
-        function Distributions.logpdf(
-            dist::LazyVectorOfUnivariate,
-            x::AbstractMatrix{<:Real},
-        )
-            size(x, 1) == length(dist) ||
-                throw(DimensionMismatch("Inconsistent array dimensions."))
-            return vec(sum(copy(logpdf.(dists, x)), dims = 1))
-        end
-
-        const LazyMatrixOfUnivariate{
-            S<:ValueSupport,
-            T<:UnivariateDistribution{S},
-            Tdists<:BroadcastArray{T,2},
-        } = MatrixOfUnivariate{S,T,Tdists}
-
-        function Distributions._logpdf(
-            dist::LazyMatrixOfUnivariate,
-            x::AbstractMatrix{<:Real},
-        )
-            return sum(copy(logpdf.(dist.dists, x)))
-        end
-
-        lazyarray(f, x...) = LazyArray(Base.broadcasted(f, x...))
-        export lazyarray
+        include("lazyarrays.jl")
     end
 end
 
--- a/src/common.jl
+++ b/src/common.jl
@@ -48,3 +48,112 @@ parameterless_type(x) = parameterless_type(typeof(x))
 parameterless_type(x::Type) = __parameterless_type(x)
 
 @non_differentiable adapt_randn(::Any...)
+
+"""
+    make_closure(f, g)
+
+Return a closure of the form `(x, args...) -> f(g(args...), x)`.
+
+# Examples
+
+This is particularly useful when one wants to avoid broadcasting over constructors
+which can sometimes cause issues with type-inference, in particular when combined
+with reverse-mode AD frameworks.
+
+```juliarepl
+julia> using DistributionsAD, Distributions, ReverseDiff, BenchmarkTools
+
+julia> const data = randn(1000);
+
+julia> x = randn(length(data));
+
+julia> f(x) = sum(logpdf.(Normal.(x), data))
+f (generic function with 2 methods)
+
+julia> @btime ReverseDiff.gradient(\$f, \$x);
+  848.759 μs (14605 allocations: 521.84 KiB)
+
+julia> # Much faster with ReverseDiff.jl.
+       g(x) = let g_inner = DistributionsAD.make_closure(logpdf, Normal)
+           sum(g_inner.(data, x))
+       end
+g (generic function with 1 method)
+
+julia> @btime ReverseDiff.gradient(\$g, \$x);
+  17.460 μs (17 allocations: 71.52 KiB)
+```
+
+See https://github.com/TuringLang/Turing.jl/issues/1934 more further discussion.
+
+# Notes
+To really go "vrooom!\" one needs to specialize on the arguments, e.g. if one
+has a function `myfunc` then we need to define
+
+```julia
+make_closure(::typeof(myfunc), ::Type{D}) where {D} = myfunc(D(args...), x)
+```
+
+This can also be done using `DistributionsAD.@specialize_make_closure`:
+
+```julia
+julia> mylogpdf(d, x) = logpdf(d, x)
+mylogpdf (generic function with 1 method)
+
+julia> h(x) = let inner = DistributionsAD.make_closure(mylogpdf, Normal)
+           sum(inner.(data, x))
+       end
+h (generic function with 1 method)
+
+julia> @btime ReverseDiff.gradient(\$h, \$x);
+  1.220 ms (37011 allocations: 1.42 MiB)
+
+julia> DistributionsAD.@specialize_make_closure mylogpdf
+
+julia> @btime ReverseDiff.gradient(\$h, \$x);
+  17.038 μs (17 allocations: 71.52 KiB)
+```
+"""
+make_closure(f, g) = (x, args...) -> f(g(args...), x)
+make_closure(f, ::Type{D}) where {D} = (x, args...) -> f(D(args...), x)
+
+
+"""
+    has_specialized_make_closure(f, g)
+
+Return `true` if there exists a specialized `make_closure(f, g)` implementation.
+"""
+has_specialized_make_closure(f, g) = false
+
+# To go vroooom we need to specialize on the first argument, thus ensuring that
+# a different closure is constructed for each method.
+"""
+    @specialize_make_closure(f)
+
+Define `make_closure` and `has_specialized_make_closure` for first first argument being `f` 
+and second argument being a type.
+"""
+macro specialize_make_closure(f)
+    return quote
+        $(DistributionsAD).make_closure(::typeof($(esc(f))), ::Type{D}) where {D} = (x, args...) -> $(esc(f))(D(args...), x)
+        $(DistributionsAD).has_specialized_make_closure(::typeof($(esc(f))), ::Type{D}) where {D} = true
+    end
+end
+
+"""
+    @specialize_make_closure(f, g)
+
+Define `make_closure` and `has_specialized_make_closure` for first first argument being `f` 
+and second argument being `g`.
+"""
+macro specialize_make_closure(f, g)
+    return quote
+        $(DistributionsAD).make_closure(::typeof($(esc(f))), ::typeof($(esc(g)))) = (x, args...) -> $(esc(f))($(esc(g))(args...), x)
+        $(DistributionsAD).has_specialized_make_closure(::typeof($(esc(f))), ::typeof{$(esc(g))}) = true
+    end
+end
+
+@specialize_make_closure Distributions.pdf
+@specialize_make_closure Distributions.logpdf
+@specialize_make_closure Distributions.loglikelihood
+@specialize_make_closure Distributions.cdf
+@specialize_make_closure Distributions.logcdf
--- a/src/lazyarrays.jl
+++ b/src/lazyarrays.jl
@@ -46,3 +46,8 @@ end
 
 lazyarray(f, x...) = BroadcastArray(f, x...)
 export lazyarray
+
+# Necessary to make `BroadcastArray` work nicely with Zygote.
+function ChainRulesCore.rrule(config::RuleConfig{>:HasReverseMode}, ::Type{LazyArrays.BroadcastArray}, f, args...)
+    return ChainRulesCore.rrule_via_ad(config, Broadcast.broadcasted, f, args...)
+end
--- a/src/lazyarrays.jl
+++ b/src/lazyarrays.jl
@@ -1,0 +1,48 @@
+using .LazyArrays: BroadcastArray, BroadcastVector, LazyArray
+
+const LazyVectorOfUnivariate{
+    S<:ValueSupport,
+    T<:UnivariateDistribution{S},
+    Tdists<:BroadcastVector{T},
+} = VectorOfUnivariate{S,T,Tdists}
+
+_inner_constructor(::Type{<:BroadcastVector{<:Any,Type{D}}}) where {D} = D
+
+function Distributions._logpdf(
+    dist::LazyVectorOfUnivariate,
+    x::AbstractVector{<:Real},
+)
+    # TODO: Implement chain rule for `LazyArray` constructor to support Zygote.
+    f = make_closure(logpdf, _inner_constructor(typeof(dist.v)))
+    # TODO: Make use of `sum(Broadcast.instantiate(Broadcast.broadcasted(f, x, args...)))` once
+    # we've addressed performance issues in ReverseDiff.jl.
+    return sum(f.(x, dist.v.args...))
+end
+
+function Distributions.logpdf(
+    dist::LazyVectorOfUnivariate,
+    x::AbstractMatrix{<:Real},
+    )
+    size(x, 1) == length(dist) ||
+        throw(DimensionMismatch("Inconsistent array dimensions."))
+    f = make_closure(logpdf, _inner_constructor(typeof(dist.v)))
+    return vec(sum(f.(x, dist.v.args...), dims = 1))
+end
+
+const LazyMatrixOfUnivariate{
+    S<:ValueSupport,
+    T<:UnivariateDistribution{S},
+    Tdists<:BroadcastArray{T,2},
+} = MatrixOfUnivariate{S,T,Tdists}
+
+function Distributions._logpdf(
+    dist::LazyMatrixOfUnivariate,
+    x::AbstractMatrix{<:Real},
+)
+    f = make_closure(logpdf, _inner_constructor(typeof(dist.v)))
+    
+    return sum(f.(x, dist.v.args))
+end
+
+lazyarray(f, x...) = BroadcastArray(f, x...)
+export lazyarray


src/DistributionsAD.jl
```diff
@@ -80,45 +80,7 @@ include("zygote.jl")
...
-            return sum(copy(logpdf.(dist.v, x)))
-        end
-
-        function Distributions.logpdf(
-            dist::LazyVectorOfUnivariate,
-            x::AbstractMatrix{<:Real},
-        )
-            size(x, 1) == length(dist) ||
-                throw(DimensionMismatch("Inconsistent array dimensions."))
-            return vec(sum(copy(logpdf.(dists, x)), dims = 1))
```
torfjelde, line 104: This even has  a bug in it: `dists` isn't defined..

src/lazyarrays.jl
```diff
@@ -0,0 +1,67 @@
+using .LazyArrays: BroadcastArray, BroadcastVector, LazyArray
```
torfjelde, line 1:
```suggestion
using .LazyArrays: BroadcastArray, BroadcastVector, LazyArray
using ChainRulesCore: ChainRulesCore
```

src/lazyarrays.jl
```diff
@@ -0,0 +1,68 @@
...
+using ChainRulesCore: ChainRulesCore
```
torfjelde, line 2:
```suggestion
using .ChainRulesCore: ChainRulesCore
```

yebai (approved):

src/common.jl
```diff
@@ -48,3 +48,112 @@ parameterless_type(x) = parameterless_type(typeof(x))
...
+        $(DistributionsAD).make_closure(::typeof($(esc(f))), ::typeof($(esc(g)))) = (x, args...) -> $(esc(f))($(esc(g))(args...), x)
+        $(DistributionsAD).has_specialized_make_closure(::typeof($(esc(f))), ::typeof{$(esc(g))}) = true
+    end
+end
+
+@specialize_make_closure Distributions.pdf
+@specialize_make_closure Distributions.logpdf
+@specialize_make_closure Distributions.loglikelihood
+@specialize_make_closure Distributions.cdf
+@specialize_make_closure Distributions.logcdf
```
devmotion, line 159: I think it should be possible to remove all of this code. Maybe type parameters are already sufficient. Or using a callable struct might help.

src/lazyarrays.jl
```diff
@@ -0,0 +1,67 @@
+using .LazyArrays: BroadcastArray, BroadcastVector, LazyArray
+
+# Necessary to make `BroadcastArray` work nicely with Zygote.
+function ChainRulesCore.rrule(config::ChainRulesCore.RuleConfig{>:ChainRulesCore.HasReverseMode}, ::Type{BroadcastArray}, f, args...)
+    return ChainRulesCore.rrule_via_ad(config, Broadcast.broadcasted, f, args...)
+end
```
devmotion, line 6: ChainRules definitions for LazyArrays.jl should really not be part of DistributionsAD. The other type piracy is already bad but at least Distributions-specific. But these lines seem really inappropriate here.

src/lazyarrays.jl
```diff
@@ -0,0 +1,67 @@
...
+    return if has_specialized_make_closure(logpdf, constructor)
+        f = make_closure(logpdf, constructor)
+        sum(f.(x, dist.v.args))
+    else
+        sum(copy(logpdf.(dist.dists, x)))
+    end
+end
+
+lazyarray(f, x...) = BroadcastArray(f, x...)
+export lazyarray
```
devmotion, line 67: It's not clear to me why this is needed. It doesn't seem much shorter and it makes it less clear that everything is based on LazyArrays.

src/common.jl
```diff
@@ -48,3 +48,112 @@ parameterless_type(x) = parameterless_type(typeof(x))
...
+  1.220 ms (37011 allocations: 1.42 MiB)
+
+julia> DistributionsAD.@specialize_make_closure mylogpdf
+
+julia> @btime ReverseDiff.gradient(\$h, \$x);
+  17.038 μs (17 allocations: 71.52 KiB)
+```
+"""
+make_closure(f, g) = (x, args...) -> f(g(args...), x)
+make_closure(f, ::Type{D}) where {D} = (x, args...) -> f(D(args...), x)
```
devmotion, line 117: I wonder if there are any possible performance/compiler benefits by not closing over the variables but to use (more-Julian) callable structs that capture `f` and `g`. In any case, I think you want
```suggestion
make_closure(f::F, g::G) where {F,G} = (x, args...) -> f(g(args...), x)
make_closure(f::F, ::Type{D}) where {F,D} = (x, args...) -> f(D(args...), x)
```

torfjelde: I unfortuantely tried that but AFAIK it ends up with this issue of closing over a `UnionAll` type again, which is exactly what we're trying to avoid (because of the issues it's causing with some AD backends) :confused: 

I might have not done it correctly though.

But you suggestion I _have_ tried, and it unfortunately doesn't have an affect. If you just look at the returned closures, they're all the same one :confused:

torfjelde: Maybe something like
```julia
struct Closure{F,G} end

Closure(::F, ::G) where {F,G} = Closure{F,G}()
Closure(::F, ::Type{G}) where {F,G} = Closure{F,G}()
Closure(::Type{F}, ::G) where {F,G} = Closure{F,G}()
Closure(::Type{F}, ::Type{G}) where {F,G} = Closure{F,G}()

for f in [pdf, logpdf, cdf, logcdf]
    @eval (::$(Closure){typeof($f),G})(x, args...) where {G} = $f(G(args...), x)
end
```
?

torfjelde:
```julia
Closure(::F, ::Type{G}) where {F,G} = Closure{F,G}()
```
and others avoid the `UnionAll` issue.

torfjelde: Have a look at what I've done now :+1:

src/lazyarrays.jl
torfjelde: Agreed! See https://github.com/JuliaArrays/LazyArrays.jl/issues/232

Think the plan is to make a glue-package for now.

Btw, this `rrule` is also not good since, for example, `logpdf(arraydist(BroadcastArray(Normal, x)), data)` will then be separated into two broadcast statements again, which is the opposite of what we want :confused: 

We _could_ of course define adjoint rules for `logpdf`, etc. that specializes on the `BroadcastArray` scenario, but that is all non-ideal :confused:

src/common.jl
```diff
@@ -48,3 +48,53 @@ parameterless_type(x) = parameterless_type(typeof(x))
...
+struct Closure{F,G} end
+
+Closure(::F, ::G) where {F,G} = Closure{F,G}()
+Closure(::F, ::Type{G}) where {F,G} = Closure{F,G}()
+Closure(::Type{F}, ::G) where {F,G} = Closure{F,G}()
+Closure(::Type{F}, ::Type{G}) where {F,G} = Closure{F,G}()
+
+@generated function (closure::Closure{F,G})(x, args...) where {F,G}
+    f = Base.issingletontype(F) ? F.instance : F
+    g = Base.issingletontype(G) ? G.instance : G
```
torfjelde, line 96: I'm not a 100% certain on this. Need to think when I've had some sleep.

src/lazyarrays.jl
devmotion: I don't think a glue package is a good long term solution. If one does not want to load ChainRulesCore for all users (even though I think it's loaded anyways in almost all realistic scenarios), a weak dependency seems the best way IMO.

src/lazyarrays.jl
devmotion: Should we deprecate it?

src/lazyarrays.jl
torfjelde: Btw got any good idea of how to define the constructor for `BroadcastArray` properly?

src/lazyarrays.jl
torfjelde: Happy to!

src/lazyarrays.jl
torfjelde: This far I have:
```julia
function ChainRulesCore.rrule(::Type{BroadcastArray}, f, args...)
    function BroadcastArray_pullback(Δ::ChainRulesCore.Tangent)
        return (ChainRulesCore.NoTangent(), Δ.f, Δ.args...)
    end
    return BroadcastArray(f, args...), BroadcastArray_pullback
end

function ChainRulesCore.rrule(
    config::ChainRulesCore.RuleConfig{>:ChainRulesCore.HasReverseMode},
    ::typeof(Distributions.logpdf),
    dist::Distributions.Product{V,D,A}, x::AbstractVector{<:Real}
) where {V,D<:Distribution,A<:BroadcastArray}
    cl = DistributionsAD.Closure(logpdf, DistributionsAD._inner_constructor(typeof(dist.v)))
    y, dy = ChainRulesCore.rrule_via_ad(config, broadcast, cl, x, dist.v.args...)
    z, dz = ChainRulesCore.rrule_via_ad(config, sum, y)

    f = dist.v.f
    function logpdf_adjoint(Δ...)
        # 1st argument is `sum` -> nothing.
        (_, sum_Δ...) = dz(Δ...)
        # 1st argument is `broadcast` -> nothing.
        # 2nd argument is `cl` -> `nothing`.
        # 3rd argument is `x` -> something.
        # Rest is `dist` arguments -> something
        (_, _, x_Δ, args_Δ...) = dy(sum_Δ...)
        # Construct the structural tangents.
        ba_tangent = ChainRulesCore.Tangent{A}(f=f, args=args_Δ)
        dist_tangent = ChainRulesCore.Tangent{typeof(dist)}(v=ba_tangent)

        return (ChainRulesCore.NoTangent(), dist_tangent, x_Δ)
    end

    return z, logpdf_adjoint
end
```
which does indeed to the trick but does not look great :confused: 

1. I don't like how I have to call back into AD, but I don't atm see a way around that.
2. I'm not too familiar with structural `Tangent`, and so I don't know if nesting them is a bad idea. For example I noticed that when we hit the pullback for `BroadcastArray`, I'm looking at a `Tangent{Any}` despite this not being the case in `logpdf_adjoint` (though this might just be Zygote doing something with it in the mean time?).
3. This isn't full support for `BroadcastArray` since in most cases it won't receive a `Tangent`.

Any more?

torfjelde: And I should probably be using `ProjectTo` instead of explicit usage of `Tangent`?

torfjelde: Awesome:) Btw, _really_ appreciate your responses over these past days; been incredibly helpful:) I'll be doing some workshop talks at a winter school on Turing next week and I would have loved to pin this down before that but we'll see :+1:

torfjelde: Btw I just pushed the resulting impl so you can look at it properly when you have time. It _does_ seem to work awfully well but it's _very_ hacky :confused:

devmotion: A PhD student from Gothenburg actually told me about the winter school and thought it might be interesting for me based on my research interests. I would have liked to go to Norway since I have only been to Bergen so far but unfortunately I don't have time (and I assume I'm not the intended audience for a workshop about Turing :stuck_out_tongue:).

src/lazyarrays.jl
```diff
@@ -0,0 +1,89 @@
...
+
+    project_broadcastarray = ChainRulesCore.ProjectTo(dist.v)
+    function logpdf_adjoint(Δ...)
+        # 1st argument is `sum` -> nothing.
+        (_, sum_Δ...) = dz(Δ...)
+        # 1st argument is `broadcast` -> nothing.
+        # 2nd argument is `cl` -> `nothing`.
+        # 3rd argument is `x` -> something.
+        # Rest is `dist` arguments -> something
+        (_, _, x_Δ, args_Δ...) = dy(sum_Δ...)
```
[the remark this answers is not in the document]
torfjelde: Okay, so that should be addressed with the `is_diff_safe` above. We will now only hit this faster path if we're certain we don't need to take derivatives wrt. anything "in" the consturctor itself.

src/common.jl
```diff
@@ -48,3 +48,93 @@ parameterless_type(x) = parameterless_type(typeof(x))
...
+false
+
+julia> # Also works on `Closure`s from `DistributionsAD`.
+       is_diff_safe(typeof(Closure(logpdf, Normal)))
+true
+
+julia> is_diff_safe(typeof(Closure(logpdf, makef([1.0]))))
+false
+"""
+@inline is_diff_safe(_) = false
```
devmotion, line 129: It seems this serves the same purpose as `ReverseDiff.mayhavetracked`: https://github.com/JuliaDiff/ReverseDiff.jl/blob/d522508aa6fea16e9716607cdd27d63453bb61e6/src/derivatives/broadcast.jl#L29

src/common.jl
```diff
@@ -48,3 +48,93 @@ parameterless_type(x) = parameterless_type(typeof(x))
...
+@inline is_diff_safe(::Type) = true
+@inline is_diff_safe(::Type{F}) where {F<:Function} = Base.issingletontype(F)
+@inline is_diff_safe(::Type{Closure{F,G}}) where {F,G} = is_diff_safe(F) && is_diff_safe(G)
+
+@generated function (closure::Closure{F,G})(x, args...) where {F,G}
+    f = Base.issingletontype(F) ? F.instance : F
+    g = Base.issingletontype(G) ? G.instance : G
+    return :($f($g(args...), x))
+end
```
devmotion, line 138: I'm not a fan of `@generated` function if there is a different way to get good performance, they have too many limitations IMO.

If we want to keep `Closure` without fields (not sure, maybe it would be better to change the heuristics in ReverseDiff), then the following seems to work at least in the example above:
```julia
julia> struct Closure{F,G} end

julia> Closure(::F, ::G) where {F,G} = Closure{F,G}()

julia> Closure(::F, ::Type{G}) where {F,G} = Closure{F,Type{G}}()

julia> Closure(::Type{F}, ::G) where {F,G} = Closure{Type{F},G}()

julia> Closure(::Type{F}, ::Type{G}) where {F,G} = Closure{Type{F},Type{G}}()

julia> (::Closure{F,G})(x, args...) where {F,G} = F.instance(G.instance(args...), x)

julia> (::Closure{F,Type{G}})(x, args...) where {F,G} = F.instance(G(args...), x)

julia> (::Closure{Type{F},G})(x, args...) where {F,G} = F(G.instance(args...), x)

julia> (::Closure{Type{F},Type{G}})(x, args...) where {F,G} = F(G(args...), x)
```
But somehow this version and the one in the PR seem all a bit hacky...

src/lazyarrays.jl
```diff
@@ -0,0 +1,99 @@
...
+
+_inner_constructor(::Type{<:BroadcastVector{<:Any,Type{D}}}) where {D} = D
+
+function Distributions._logpdf(
+    dist::LazyVectorOfUnivariate,
+    x::AbstractVector{<:Real},
+)
+    # TODO: Make use of `sum(Broadcast.instantiate(Broadcast.broadcasted(f, x, args...)))` once
+    # we've addressed performance issues in ReverseDiff.jl.
+    constructor = _inner_constructor(typeof(dist.v))
```
devmotion, line 17: I'm sure this will be problematic in some cases and break. It's not guaranteed that `_inner_constructor` returns a proper constructor. Something safer would be https://github.com/JuliaObjects/ConstructionBase.jl I assume.

devmotion: A simple example:
```julia
julia> struct A{X,Y}
           x::X
           y::Y
           A(x::X, y::Y) where {X,Y} = new{X,Y}(x, y)
       end

julia> _constructor(::Type{D}) where {D} = D
_constructor (generic function with 1 method)

julia> x, y = 1, 2.0
(1, 2.0)

julia> a = A(x, y)
A{Int64, Float64}(1, 2.0)

julia> _constructor(typeof(a))(x, y)
ERROR: MethodError: no method matching A{Int64, Float64}(::Int64, ::Float64)
Stacktrace:
 [1] top-level scope
   @ REPL[31]:1
```

src/lazyarrays.jl
```diff
@@ -0,0 +1,99 @@
...
+    function BroadcastArray_pullback(Δ::ChainRulesCore.Tangent)
+        return (ChainRulesCore.NoTangent(), Δ.f, Δ.args...)
+    end
+    return BroadcastArray(f, args...), BroadcastArray_pullback
+end
+
+ChainRulesCore.ProjectTo(ba::BroadcastArray) = ProjectTo{typeof(ba)}((f=ba.f,))
+function (p::ChainRulesCore.ProjectTo{BA})(args...) where {BA<:BroadcastArray}
+    return ChainRulesCore.Tangent{BA}(f=p.f, args=args)
+end
```
devmotion, line 60: I'm surprised this is needed. Feels like that's the default for `Tangent`s anyway?

src/lazyarrays.jl
```diff
@@ -0,0 +1,99 @@
...
+
+function ChainRulesCore.rrule(
+    config::ChainRulesCore.RuleConfig{>:ChainRulesCore.HasReverseMode},
+    ::typeof(Distributions.logpdf),
+    dist::LazyVectorOfUnivariate,
+    x::AbstractVector{<:Real}
+)
+    # Extract the constructor used in the `BroadcastArray`.
+    constructor = DistributionsAD._inner_constructor(typeof(dist.v))
```
devmotion, line 69: As mentioned above, it's not guaranteed that this actually is a constructor that can be called with the arguments.

devmotion: BTW is the qualification really needed?

src/lazyarrays.jl
```diff
@@ -0,0 +1,99 @@
...
+end
+
+ChainRulesCore.ProjectTo(ba::BroadcastArray) = ProjectTo{typeof(ba)}((f=ba.f,))
+function (p::ChainRulesCore.ProjectTo{BA})(args...) where {BA<:BroadcastArray}
+    return ChainRulesCore.Tangent{BA}(f=p.f, args=args)
+end
+
+function ChainRulesCore.rrule(
+    config::ChainRulesCore.RuleConfig{>:ChainRulesCore.HasReverseMode},
+    ::typeof(Distributions.logpdf),
```
devmotion, line 64:
```suggestion
    ::typeof(logpdf),
```

src/lazyarrays.jl
```diff
@@ -0,0 +1,99 @@
...
+    dist::LazyVectorOfUnivariate,
+    x::AbstractVector{<:Real}
+)
+    # Extract the constructor used in the `BroadcastArray`.
+    constructor = DistributionsAD._inner_constructor(typeof(dist.v))
+
+    # If it's not safe to ignore the `constructor` in the pullback, then we fall back
+    # to the default implementation.
+    is_diff_safe(constructor) || return ChainRulesCore.rrule_via_ad(config, (d,x) -> sum(logpdf.(d.v, x)), dist, x)
```
devmotion, line 73: Could we just use
```suggestion
    is_diff_safe(constructor) || return ChainRulesCore.rrule_via_ad(config, (d,x) -> logpdf(d, x), dist, x)
```
to avoid making any assumptions about how `logpdf(dist, x)` is implemented?

devmotion: Alternatively maybe just return `nothing`?
```suggestion
    is_diff_safe(constructor) || return nothing
```
(https://juliadiff.org/ChainRulesCore.jl/stable/ad_author/opt_out.html)

src/lazyarrays.jl
```diff
@@ -0,0 +1,99 @@
...
+    end
+    return BroadcastArray(f, args...), BroadcastArray_pullback
+end
+
+ChainRulesCore.ProjectTo(ba::BroadcastArray) = ProjectTo{typeof(ba)}((f=ba.f,))
+function (p::ChainRulesCore.ProjectTo{BA})(args...) where {BA<:BroadcastArray}
+    return ChainRulesCore.Tangent{BA}(f=p.f, args=args)
+end
+
+function ChainRulesCore.rrule(
```
devmotion, line 62: BTW the annoying thing about these kinds of general rules is that it might break (and it happened to me multiple times) code that would have worked without rule and if one would just let the AD system perform its default differentiation. One can fix these issues though by using `@opt_out` (but that also has some problems...).

src/lazyarrays.jl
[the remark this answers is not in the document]
devmotion: The following example in ChainRules is related: https://github.com/JuliaDiff/ChainRules.jl/blob/9adf759bc63432dc518ccf499d6938fc5a217113/src/rulesets/Base/mapreduce.jl#L76

src/lazyarrays.jl
```diff
@@ -0,0 +1,99 @@
...
+    cl = DistributionsAD.Closure(logpdf, constructor)
+
+    # Construct pullbacks manually to avoid the constructor of `BroadcastArray`.
+    y, dy = ChainRulesCore.rrule_via_ad(config, broadcast, cl, x, dist.v.args...)
+    z, dz = ChainRulesCore.rrule_via_ad(config, sum, y)
+
+    project_broadcastarray = ChainRulesCore.ProjectTo(dist.v)
+    function logpdf_adjoint(Δ...)
+        # 1st argument is `sum` -> nothing.
+        (_, sum_Δ...) = dz(Δ...)
```
devmotion, line 85: Generally you might have to deal with `unthunk` I assume.

devmotion (commented):
I had a rather quick look. There's many things that should not be here (but maybe some them could be moved or fixed somewhere else). Specializations for LazyArrays could maybe be added to Distributions as weak dependencies (or maybe we just fix the use of `Broadcasted`). Nevertheless it might be OK for some time to have fixes here, but we should make sure to test everything.

src/common.jl
torfjelde: Yeah basically, though I was looking at Zygote's `_dual_purefun` because ReverseDiff already hits the `broadcast_forward`, even without the custom adjoint I defined (ReverseDiff also doesn't support calling back into AD).

src/common.jl
torfjelde: > But somehow this version and the one in the PR seem all a bit hacky...

As mentioned before, I 100% agree with you. _But_ this performance issue is literally the cause of several Slack and Discourse threads of people going "why is Turing so slow for this simple model?", and so IMO we should just get this fixed despite its hackiness and then we make it less hacky as  we go + maybe improve ReverseDiff and Zygote.

torfjelde: Oh and regarding the `@generated`, I'm happy to do away with it. I'll try your suggestion :+1:

src/lazyarrays.jl
torfjelde: So I was actually using ConstructionBase locally for this before:) But I removed it because I figured this will only be used for a very simple subset of constructors, so uncertain if it's worth it. But I'll add it back again:)

src/lazyarrays.jl
torfjelde: Yeah, so we can alos just close over the function `f` and construct the `Tanget` directly in the adjoint (in fact, this is what I did originally), but I thought maybe `ProjectTo` was the more "proper" way to do it. I can go back to directly constructing the `Tangent` though:)

src/common.jl
```diff
@@ -48,3 +48,93 @@ parameterless_type(x) = parameterless_type(typeof(x))
...
+```
+
+See https://github.com/TuringLang/Turing.jl/issues/1934 more further discussion.
+"""
+struct Closure{F,G} end
+
+Closure(::F, ::G) where {F,G} = Closure{F,G}()
+Closure(::F, ::Type{G}) where {F,G} = Closure{F,G}()
+Closure(::Type{F}, ::G) where {F,G} = Closure{F,G}()
+Closure(::Type{F}, ::Type{G}) where {F,G} = Closure{F,G}()
```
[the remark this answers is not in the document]
torfjelde: So I originally had fields but yeah this resulted in bad computation paths. Might be something that should be changed in the AD instead, true.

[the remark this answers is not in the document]
torfjelde: Worth pointing out that
1. `Closure` is not something that's meant to be used heavily for arbitrary callables (e.g. the checks in the adjoint explicitly exclude the scenario where we have fields).
2. `Closure` should not be used by the end-user.

src/lazyarrays.jl
torfjelde: > to avoid making any assumptions about how logpdf(dist, x) is implemented?

But won't this hit the same `rrule` once you get to `logpdf`? 

> Alternatively maybe just return nothing?

I don't completely understand this. So opting out means it will fall back to the underlying AD? I thought it meant "nothing to differentiate here"

src/lazyarrays.jl
torfjelde: We could also just define this using `ZygoteRules.@adjoint` if that helps? 

Unfortunately there's no way around this because we have to stop Zygote from trying to differentiate through the broadcasted constructor.

src/lazyarrays.jl
torfjelde: But should the other pullbacks also deal with this?

torfjelde: Seems like we only need to deal with `unthunk` in the pullback for the `BroadcastArray` constructor, no?

torfjelde:
> I had a rather quick look. There's many things that should not be here (but maybe some them could be moved or fixed somewhere else). Specializations for LazyArrays could maybe be added to Distributions as weak dependencies (or maybe we just fix the use of Broadcasted). Nevertheless it might be OK for some time to have fixes here, but we should make sure to test everything.

I 100% agree with you, but realistically tackling these issues is going to take a lot of time and effort. In the mean time, I think this hacky approach will have to do :confused: This solves so many Slack and Discourse threads of people going "why is this simple model so slow in Turing.jl?"...

svilupp:
I was wondering what is stopping this PR from being merged? 
I've had this page in my open tabs for a few weeks now, anxiously awaiting the purple merged badge 😄 

I was hoping to re-do my logistic regression benchmark with this PR once it's merged, but I jumped the gun [here](https://discourse.julialang.org/t/case-study-speeding-up-a-logistic-regression-with-rhs-prior-turing-vs-numpyro-any-tricks-im-missing/87681/37?u=svilupp).
  
In short, the benefits are incredible and the best part is that it would be easy even for new users (it just needs an update in the tutorials).

src/DistributionsAD.jl
```diff
@@ -70,6 +70,13 @@ include("zygote.jl")
         end
     end
 
+    @require Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f" begin
+        using .Zygote: Zygote
+        # HACK: Make Zygote (correctly) recognize that it should use `ForwardDiff` for broadcasting.
+        # See `is_diff_safe` for more information.
+        Zygote._dual_purefun(::Type{C}) where {C<:Closure} = is_diff_safe(C)
+    end
```
ChrisRackauckas, line 78: Make into a weak dep?

devmotion: Haha yeah that should be done with basically everything in this repo. Or rather, most things should be moved to Distributions and added as a weak dep there to fix the type piracy issues.

ToucheSir: Ideally `is_diff_safe` would be a function ForwardDiff or one of its dependencies owns. That would avoid the need for a package extension or requires block overriding an internal function like this.

torfjelde:
> I was wondering what is stopping this PR from being merged?

> I've had this page in my open tabs for a few weeks now, anxiously awaiting the purple merged badge smile

The approach taken here is quite hacky, and, tbh, it's unfortunate that we even have to do this. It's really just working around type-inference issues for calling a `UnionAll`. 

So if we accept, we have to be on high-alert in case something breaks (due to its hackinesse), and ideally this should be fixed elsewhere so maybe we should spend some more time thinking to see if we can address this in a broader manner.

> I was hoping to re-do my logistic regression benchmark with this PR once it's merged, but I jumped the gun here.

> In short, the benefits are incredible and the best part is that it would be easy even for new users (it just needs an update in the tutorials).

Glad to see it's working though!:)

svilupp:
> > I was wondering what is stopping this PR from being merged?
> 
> > I've had this page in my open tabs for a few weeks now, anxiously awaiting the purple merged badge smile
> 
> The approach taken here is quite hacky, and, tbh, it's unfortunate that we even have to do this. It's really just working around type-inference issues for calling a `UnionAll`.
> 
> So if we accept, we have to be on high-alert in case something breaks (due to its hackinesse), and ideally this should be fixed elsewhere so maybe we should spend some more time thinking to see if we can address this in a broader manner.
> 
> > I was hoping to re-do my logistic regression benchmark with this PR once it's merged, but I jumped the gun here.
> 
> > In short, the benefits are incredible and the best part is that it would be easy even for new users (it just needs an update in the tutorials).
> 
> Glad to see it's working though!:)

Since this PR is unlikely to be merged, do we have tips/snippets that intermediate users could opt into in their code?

**How would I recognize that the UnionAll is slowing me down (besides the slowness itself)?**
I can see that you identified it with [this call](https://github.com/TuringLang/Turing.jl/issues/1934#issuecomment-1382308140) by running `@code_warntype f.makeargs.f` and looking for `Body::Any`.

However, it's not clear to me "where" to break into. 
How would I set a breakpoint for it with Debugger?
```
using Debugger, Turing
# as per your example
@model function irt(y, i, p; I = maximum(i), P = maximum(p))
    theta ~ filldist(Normal(), P)
    beta ~ filldist(Normal(), I)
    Turing.@addlogprob! sum(logpdf.(BernoulliLogit.(theta[p] - beta[i]), y))

    return (; theta, beta)
end
model = irt(y, i, p); 
# what is the `some_func` at which we should break to see the context
@bp some_func
@run chn=sample(model, NUTS(), 100)
```

**In the absence of this PR, what's the best way to overcome the AD taking the slow path?**
My understanding of your thread was that it comes from broadcasting over Structs, so we want compiler to remove them.

Eg, define a wrapper that is a function
```
mylogpdf(logit,y) = logpdf(BernoulliLogit(logit),y)

# to be used like
Turing.@addlogprob! sum(mylogpdf.(theta[p] - beta[i], y))

# instead of this in Turing.@addlogprob! 
Turing.@addlogprob! sum(logpdf.(BernoulliLogit.(theta[p] - beta[i]), y))
```
and check as per the above point if you get `Body::Any` or a type

Or would you go about it differently?

PR #235, opened 2023-04-26
Switch to extensions

### LINKED ISSUE #234: Try change to using extensions for improved load time.
In profiling some packages (for example OmniPackage.jl) it shows that a significant time is spent in `__init__` in this package (2.5s). It is likely that this package would be significantly faster to load if it used extensions. It would be a worthwhile trial imo to switch to that.

It also seems that some packages that are unconditionally loaded (and thus not use Requires) could technically also use extensions, for example Zygote.

cc @devmotion

devmotion:
Fixes #234.


devmotion:
There's a surprising but seemingly reproducible test failure on Linux with ForwardDiff AD + Julia 1.9. The error message is
```
signal (15): Terminated
in expression starting at /home/runner/work/_actions/julia-actions/julia-runtest/latest/test_harness.jl:7
```

Otherwise it seems this PR does not break any tests.

torfjelde:
Just saw this! Will have a look later tonight or tomorrow morning!

torfjelde (approved):
LGTM! Two questions: have the issue with ForwardDiff mentioned above been fixed (tests are passing?), and version bump?

devmotion:
> have the issue with ForwardDiff mentioned above been fixed (tests are passing?

No, the problematic ForwardDiff configuration (Ubuntu + Julia 1.9) still failed; only the other configurations passed. Let's see if this happens again with the latest commit.

> version bump?

I bumped the version (initial bump became outdated due to the StatsBase update). I also cleaned the imports a bit (seems e.g. we don't need to depend on StatsBase and NaNMath anymore).

> Would be interesting to see if this happens on the master branch of this package with the same set of versions used here.

Good idea, I opened https://github.com/TuringLang/DistributionsAD.jl/pull/243 to check master.

devmotion:
Would be good to fix them but then I'm very confident that it's unrelated to this PR. Since ForwardDiff tests pass on Julia 1.6 and MacOS + Julia 1.9, I'll merge and release.

devmotion:
I see similar errors on Ubuntu + Julia 1.9 in the ReverseDiff repo: https://github.com/JuliaDiff/ReverseDiff.jl/actions/runs/4967162516/jobs/8889466679#step:6:637

PR #245, opened 2023-06-06
LazyArrays.jl extension is broken

torfjelde:
Extension for LazyArrays is using `ValueSupport` without having imported it. This fixes that.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "DistributionsAD"
 uuid = "ced4e74d-a319-5a8a-b0ac-84af2272839c"
-version = "0.6.47"
+version = "0.6.48"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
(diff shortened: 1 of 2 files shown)


devmotion (commented):
Could we add a test to ensure that this extension works (or can be loaded at least)?

torfjelde:
What should that be? Just loading the package in the tests?

Btw thoughts on the breaking Tracker tests? Are fixing these or just marking as broken?

devmotion:
If we don't load it in the tests, then yes  we should add least do that. I guess that might have caught the bug?

This is not a Tracker issue, a recent release of FillArrays broke it. I already opened an issue upstream and some bugs were fixed, but unfortunately not all of them it seems.

torfjelde:
> If we don't load it in the tests, then yes we should add least do that. I guess that might have caught the bug?

Yeah, that would have caught it.

> This is not a Tracker issue, a recent release of FillArrays broke it. I already opened an issue upstream and some bugs were fixed, but unfortunately not all of them it seems.

Sooooo whadda we do? Related to https://github.com/TuringLang/Turing.jl/issues/2000 :eyes: 

_But_ Tracker tests are also breaking on master, so I don't think this should hold this PR back, as this PR is really fixing a bug + changes should in no way affect any of the current tests. So can we at least merge this regardless of the Tracker-issues?

devmotion:
Yes, I agree, the Tracker issue should not hold back this PR. But it seems we still have to add a test for the fix?

PR #246, opened 2023-06-10
Add regression test for LazyArrays

devmotion:
@torfjelde Would that have been sufficient for catching #245?


torfjelde:
Yup! This would have done it:) Sorry about the late reply.

torfjelde (approved):

PR #247, opened 2023-06-11
Actual fix for LazyArrays extension

torfjelde:
_Now_ it should be working.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "DistributionsAD"
 uuid = "ced4e74d-a319-5a8a-b0ac-84af2272839c"
-version = "0.6.48"
+version = "0.6.49"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
--- a/ext/DistributionsADLazyArraysExt.jl
+++ b/ext/DistributionsADLazyArraysExt.jl
@@ -3,12 +3,12 @@ module DistributionsADLazyArraysExt
 if isdefined(Base, :get_extension)
     using DistributionsAD
     using LazyArrays
-    using DistributionsAD: Distributions, ValueSupport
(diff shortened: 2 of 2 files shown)


torfjelde:
I will say, all these broken tests (that are unrelated to this PR) makes it very difficult to tell whether the PR introduces any bugs :confused:

devmotion:
Maybe we should just upper bound the FillArrays compat of all existing and future releases until we know what in the recent FillArrays release broke it.

The PR seems fine but can we extend our tests for the extension? Seems the ones in #246 were not sufficient.

torfjelde:
I'm very confused as to what's going on the tests of #246 though. If you look at the logs for some of the tests, e.g. https://github.com/TuringLang/DistributionsAD.jl/actions/runs/5230268767/jobs/9443723459, you can see that it's indeed erroring on that important (just warning in the case of Revise.jl). But then on some tests it's still passing https://github.com/TuringLang/DistributionsAD.jl/actions/runs/5237692580/jobs/9456041448?pr=247.

devmotion:
Ah, I had missed that some tests had failed due to these bugs. But yeah, strange that it doesn't fail every time.

torfjelde:
> Ah, I had missed that some tests had failed due to these bugs.

Same; that's what trigger my comment about the currently failing tests makes it a bit harder to see what's going wrong :sweat_smile: 

> Maybe we should just upper bound the FillArrays compat of all existing and future releases until we know what in the recent FillArrays release broke it.

Do we have any idea if this will have downstream consequences? :Confused

torfjelde:
But can we just get this merged? It's holding up quite a few PRs? :confused: I feel like upper-bounding FillArrays is a separate topic anyway.

torfjelde:
This is the same as before with our TRacker support being broken, which was the case prior to this PR too (@devmotion is saying it's because of some FillArrays stuff, hence why he suggested we might want to upper bound FIllArryas for now)

PR #248, opened 2023-06-13
Try to fix tests by bounding FillArrays

torfjelde:


torfjelde:
This seems to have done it? :grimacing: I've only restricted the tests though, not the package deps

torfjelde:
Okay, so this does indeed fix the reverse-mode issues, but it's soooooo slow. 

This might be due to some stuff @devmotion mentioned. I see there have been some performance fixes after FillArrays@1.0.0.

devmotion:
Supposed to be fixed by FillArrays 1.4.0.

PR #251, opened 2023-07-11
Update FillArrays compat to 1.4.0

devmotion:
Closes #248.


devmotion:
Seems the final commit that fixed type inference broke ReverseDiff: https://github.com/TuringLang/DistributionsAD.jl/pull/251 AD and broadcasting is.... annoying.

devmotion:
Possibly fixed by https://github.com/JuliaArrays/FillArrays.jl/pull/278, let's see.

PR #256, opened 2023-08-04
Fix missing `logpdf`

yebai:
Fix [errors](https://github.com/TuringLang/TuringGLM.jl/actions/runs/5284783602/job/14543012171?pr=92#step:6:1360) observed in `TuringGLM`.


storopoli:
Thanks, I've been trying to find the source of the errors for the past weeks with no success...

storopoli (approved):

devmotion (approved):
Would be good to have some tests that would have caught the issue, the lazy extension is severely under-tested. Maybe we should open an issue?

In any case, the fixes seem good.

PR #258, opened 2023-09-24
Use `twoπ` and convert it

### LINKED ISSUE #156: Unwanted type promotion from Float32 to Float64 while calculating logpdf of TuringDiagMvNormal
While computing logpdf of Multivariate normal distributions, the logpdf is of type Float64, even when the mean, covariance matrix and observation all have eltype Float32.
MVE for TuringDiagMvNormal (TuringScalMvNormal and TuringDenseMvNormal suffer the same problem):
```julia
julia> using Distributions, DistributionsAD
julia> d = TuringDiagMvNormal(zeros(Float32, 2), ones(Float32, 2))
TuringDiagMvNormal{Vector{Float32}, Vector{Float32}}(m=Float32[0.0, 0.0], σ=Float32[1.0, 1.0])
julia> l = logpdf(d, [1f0, 2f0])
-4.337877066409345
julia> typeof(l)
Float64  #expected Float32
```
After a quick glance at the source, the problem seems to be that the constant 2π is Float64, which then promotes the rest of the expression.

prbzrg:
Fixes #156


prbzrg:
I used `IrrationalConstants.twoπ` and removed `eltype` definitions.

src/multivariate.jl
```diff
@@ -154,40 +154,46 @@ end
...
 end
 function Distributions.loglikelihood(d::TuringScalMvNormal, x::AbstractMatrix{<:Real})
     σ2 = abs2(d.σ)
-    return -(length(x) * log(2π * σ2) + sum(abs2.(x .- d.m)) / σ2) / 2
+    return -(length(x) * log(twoπ * σ2) + sum(abs2.(x .- d.m)) / σ2) / 2
 end
 
 function Distributions._logpdf(d::TuringDiagMvNormal, x::AbstractVector)
-    return -(length(x) * log(2π) + 2 * sum(log.(d.σ)) + sum(abs2.((x .- d.m) ./ d.σ))) / 2
+    s = sum(log, d.σ)
```
devmotion, line 170: sum(log....) to sum(log, ...)) which in the past caused sometimes problems with AD. Better not include such changes.

devmotion:
Closed in favour of #259 to avoid unrelated changes.

devmotion:
Thank you for the PR but I think it's better to keep it as minimal as possible 🙂

prbzrg:
OK, thanks for reviewing.
If there is anything I can do to help, I would like to participate.

PR #259, opened 2023-09-24
Fix promotions with irrationals

### LINKED ISSUE #156: Unwanted type promotion from Float32 to Float64 while calculating logpdf of TuringDiagMvNormal

devmotion:
This PR shows what I had in mind when commenting on #258.

Should fix #156 but is quickly typed on my mobile phone and untested.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "DistributionsAD"
 uuid = "ced4e74d-a319-5a8a-b0ac-84af2272839c"
-version = "0.6.52"
+version = "0.6.53"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
--- a/src/multivariate.jl
+++ b/src/multivariate.jl
@@ -154,40 +154,46 @@ end
 
 function Distributions._logpdf(d::TuringScalMvNormal, x::AbstractVector)
     σ2 = abs2(d.σ)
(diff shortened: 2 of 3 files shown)


prbzrg:
There's a way that already exist:
https://github.com/TuringLang/DistributionsAD.jl/blob/5847e86f7783ea8f745a7465c2f4f9020c729051/src/multivariate.jl#L198-L199

Thanks for the idea, I updated my PR. my goal is the issue to be fixed; feel free to close my PR.

devmotion:
I'm currently AFK but I'll finalize the PR (and eg add tests) tonight. Then we should be able to merge and release the fix within the next few days.

PR #261, opened 2024-01-31
Bump compat for Adapt to 4

fhagemann:
Adapt v4.0.0 features breaking changes in `Adapt.ndims`, `Adapt.eltype` and `Adapt.parent`
(see https://github.com/JuliaGPU/Adapt.jl/releases/tag/v4.0.0)

DistributionsAD.jl seems to be using only `Adapt.adapt` [here](https://github.com/TuringLang/DistributionsAD.jl/blob/b7c2184b7791fa5f379ef3948e760961890eb40f/src/common.jl#L42), so it should not be affected by the changes in Adapt v4.0.0.

Closes #262.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -31,7 +31,7 @@ DistributionsADReverseDiffExt = "ReverseDiff"
 DistributionsADTrackerExt = "Tracker"
 
 [compat]
-Adapt = "2, 3"
+Adapt = "2, 3, 4"
 ChainRules = "1.35.3"
 ChainRulesCore = "1"
 Compat = "3.6, 4"


avik-pal:
@devmotion can this be merged and tagged? It is holding back a downgrade CI

devmotion:
I'll merge if tests pass.

> It is holding back a downgrade CI

How is that possible? Doesn't a test failure of downgrade CI just indicate that the compat bounds of your package have to be fixed?

avik-pal:
Had to drop compat with Adapt 3 at one place, and this is the only one currently in the dependency chain with no Adapt 4 compat

devmotion:
I still don't understand, how can you drop Adapt 3 if it is not supported yet by DistributionsAD? Shouldn't then the regular package tests, installation etc. fail as well?

avik-pal:
The compat has not yet been dropped. I was running downgrade CI and after running through a few combinations this is only one which doesn't have Adapt 4, and then testing locally with this branch the dependency resolution works out

PR #264, opened 2024-04-15
Added default impls for `filldist` and `arraydist`

torfjelde:
Currently we don't support something like `filldist(Dirichlet(ones(2)), 3, 4)` even though there is a distribution impl available in Distributions.jl through `product_distribution`.

This PR adds such default impls for high-dim calls to `filldist` and `arraydist`; AD might be worse performance-wise but we should still support it.

https://github.com/TuringLang/Turing.jl/issues/2190

### DIFF
--- a/src/arraydist.jl
+++ b/src/arraydist.jl
@@ -1,3 +1,10 @@
+"""
+    arraydist(dists)
+
+Create a distribution from an array of distributions.
+"""
+arraydist(dists::AbstractArray) = product_distribution(dists)
+
 # Univariate
 
 const VectorOfUnivariate = Distributions.Product
--- a/src/filldist.jl
+++ b/src/filldist.jl
@@ -1,5 +1,14 @@
+# Default implementation just defers to Distributions.jl.
+"""
+    filldist(d::Distribution, ns...)
+
+Create a product distribution using `FillArrays.Fill` as the array type.
+"""
+filldist(d::Distribution, ns...) = product_distribution(Fill(d, ns...))
+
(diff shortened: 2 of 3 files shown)


torfjelde:
@yebai @devmotion should be quick merge

src/arraydist.jl
```diff
@@ -1,3 +1,10 @@
+"""
+    arraydist(dists)
+
+Create a distribution from an array of distributions.
+"""
+arraydist(dists::AbstractArray) = product_distribution(dists)
```
devmotion, line 6: Should we limit this fallback to
```suggestion
arraydist(dists::AbstractArray{<:Distribution}) = product_distribution(dists)
```
?

src/filldist.jl
```diff
@@ -1,5 +1,14 @@
+# Default implementation just defers to Distributions.jl.
+"""
+    filldist(d::Distribution, ns...)
+
+Create a product distribution using `FillArrays.Fill` as the array type.
+"""
+filldist(d::Distribution, ns...) = product_distribution(Fill(d, ns...))
```
devmotion, line 7: Maybe require at least one dimension?
```suggestion
filldist(d::Distribution, n1::Int, ns::Int...) = product_distribution(Fill(d, n1, ns...))
```

devmotion:
Maybe add (non-AD?) tests?

torfjelde:
Ah yes good point!

torfjelde:
You happy with it now @devmotion ?

PR #272, opened 2024-11-03
Don't check result of `rand(d)` in doc tests

devmotion:
Checking the output of `rand(d)` in a doc test is very brittle, even when setting a seed or even with StableRNGs. For instance, https://github.com/JuliaStats/Distributions.jl/pull/1905 breaks this doctest.


penelopeysm (approved):
Thanks!

PR #279, opened 2025-03-30
ForwardDiff -> 1

penelopeysm:
supersedes #277, #278

minor complaint about how compathelper is annoyingly unhelpful because it bumps the compat in Project.toml -> tests run fine but they don't use the new version because test/Project.toml isn't updated. and then in a separate PR it bumps test/Project.toml but doesn't bump main compat so test env fails to resolve


penelopeysm:
The Zygote failure is, apparently, because I downgraded CI to 'min' rather than '1.6'.

It turns out that Zygote fails to precompile on anything less than 1.6.5. In an ideal world, this would be fixed upstream by yanking all Zygote releases that claimed to work on >= 1.6 but didn't. Since our world is non-ideal, I'm going to bump the minimum compat here instead.

